PR #84, opened 2015-06-09
Update /po/zh_CN.po

dainslef:
Update some Cinese translations.


flexiondotorg:
Thanks for the contribution, but Translations should be submitted via Transifex :-)

  * https://www.transifex.com/organization/mate/dashboard/MATE

PR #86, opened 2015-06-21
Update README

ZrN-kAyDqY:
* Notes

  [Read-only mode] means that Engrampa may read the file type, but it cannot write to that file type.  
  An example is ISO-9660 CD Disc Image (.iso) [Read-only mode].  ISO images may be read, but
  file(s) cannot be written to an (.iso).

### DIFF
--- a/README
+++ b/README
@@ -79,3 +79,9 @@
   ./autogen.sh
   make
   make install
+
+* Notes
+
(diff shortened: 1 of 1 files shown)


ZrN-kAyDqY:
Clarifies that this news item for 1.10 "allow to compress iso images and tar archives from the caja menu"  does not mean that .iso may be created with the tool.

PR #98, opened 2015-09-17
fixed ngettext usage

mikhirev:
Hi!
I found a mistake in ngettext usage while working on translation. Please merge the fix and update pot/po files.


monsta:
Corrected it a bit to match [upstream commit](https://git.gnome.org/browse/file-roller/commit?id=9156e6e8f3955d1a35491658315e1df06b0355a2) and merged in https://github.com/mate-desktop/engrampa/commit/2c856871591b413872b11f48bd725d71b8ebbc2a, thanks :smile:

As for new translations, they're handled at Transifex, so we'll need to push these changes there, wait for the translators to do their work, and sync the translations from there.

Issue #78, opened 2015-02-02
Cannot extract mulpiple files concurently to a seprate directory


mf0hka:
I had 18 zip files to decompress, I wanted to move the uncompressed folders to a sepreate directory. 
when i ran the decompress to a specific location, only 1 of the 18 decompressed properly. 

So then I tried highlighting all of the compressed packages, and chose "extract here". All the files and folders were promptly decompressed. so that works. 

So then I tried it with 2 zipped packages. That extracted to a specified directory as expected.
Next I tired extracting 10 zipped packages to a directory spcified, and that failed. By fail, I mean the operation did not produce the expected output. There have been no message boxes or errors, or warnings that I have seen. 

Then I tried decompressing 5 packages to a directory. That failed. Only the 5th package was extracted to the selected directory.. 

I want to note, my zip package names are php01 through php18
Im going to sleep now. Iam using debian 8 with mate 1.8

sc0w:
I solved this issue:

https://github.com/mate-desktop/engrampa/pull/100

The patch:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=800817

PR #100, opened 2015-10-03
Fixes: Cannot extract multiple files concurrently to a separate directory

sc0w:
Fixes https://github.com/mate-desktop/engrampa/issues/78

This is a new feature on engrampa :)

The Patch:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=800817


raveit65:
works like a charm, thank you.

PR #111, opened 2015-11-21
Let's add some parallelism to engrampa

oz123:
I'd like to add support for parallel programs in engrampa. Most of the utils working with tar have a mulitcore implementation which is faster. This really cool if you can do in the command line the following:
``` 
tar -Ipixz -xvf file.tar.xz 
```
instead of 
```
tar xvJf file.tar.xz
```
The first command will utilize all cores on the machine and will run significantly faster.
Most machines sold today (and in the last years even) are multi-core, even simple boards like the raspberrypi. So why limit only CLI users and power users to that? 

This patch allows to build engrampa with the following option:
```
./configure --enable-pixz
make
make install
```
Now files with xz compression will be extracted with pixz instead of xz binary.

I intend to add a series of patches to enable parallel extraction to all the possible backends (read: where a good free implementation is packaged and found). 

This would be a humble start. After that we could think of : 
- parallel compressing,
- modifying the GUI to enable fine control (e.g. max CPUS used for extraction\compression)
- instead of build time option, user (not packager) ability to choose between parallel and conventional compression program (pixz\xz for example). 

I would be happy to hear your feedback about this. If the feedback is positive, I will go a head an add the other parallel options to the code (e.g. bz2, gunzip etc.)

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -4,12 +4,9 @@
   an entry in po/ChangeLog
 
 *Patches
-
-  Send patches to the maintainer (engrampa-maint@gnome.org) and get 
-  approval before committing.
-
-  Patches must contain a ChangeLog entry.
-
+  
+  Fork and send Pull Requests to github.com/mate-desktop/enrampa.
+  
   You have to follow the style of the rest of the code even if you 
   don't like it.  The code style is K&R with 8 space tabs.
 
--- a/HACKING
+++ b/HACKING
@@ -5,7 +5,7 @@
 
 *Patches
   
-  Fork and send Pull Requests to github.com/mate-desktop/enrampa.
+  Fork and send Pull Requests to github.com/mate-desktop/engrampa.
   
   You have to follow the style of the rest of the code even if you 
   don't like it.  The code style is K&R with 8 space tabs.
--- a/configure.ac
+++ b/configure.ac
@@ -217,6 +217,22 @@ AM_CONDITIONAL(ENABLE_MAGIC, test x"$enable_magic" != x"no")
 
 dnl ******************************
 
+dnl ******************************
+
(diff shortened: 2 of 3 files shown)


sc0w:
hello

I have tested,  and really pixz is more quikly with big files

but, with your changes on configure.ac , If the user install engrampa whithout pixz, and then install pixz, engrampa uses xz

I think this is better instead of configure.ac, and it uses pixz when it is installed on the system:

@oz123 what do you think about this?

```patch
diff --git a/src/fr-command-tar.c b/src/fr-command-tar.c
index d52964c..3a71410 100644
--- a/src/fr-command-tar.c
+++ b/src/fr-command-tar.c
@@ -216,7 +216,11 @@ add_compress_arg (FrCommand *comm)
 		fr_process_add_arg (comm->process, "--use-compress-program=lzma");
 
 	else if (is_mime_type (comm->mime_type, "application/x-xz-compressed-tar"))
-		fr_process_add_arg (comm->process, "--use-compress-program=xz");
+		if (is_program_in_path ("pixz")) {
+			fr_process_add_arg (comm->process, "--use-compress-program=pixz");
+		} else {
+			fr_process_add_arg (comm->process, "--use-compress-program=xz");
+		}
 
 	else if (is_mime_type (comm->mime_type, "application/x-lzop-compressed-tar"))
 		fr_process_add_arg (comm->process, "--use-compress-program=lzop");

oz123:
@sc0w not bad either! Glad to see someone cares about this patch after such a long time.

sc0w:
thanks

https://github.com/mate-desktop/engrampa/commit/d9fa51d3a2a88e0b800bfc989ca285f1a6125f69

oz123:
YAY!

PR #117, opened 2016-01-04
p7zip v9.38 or later with arg -spd

### LINKED ISSUE #108: p7zip and problems with char '?' on a compressed file
with p7zip 9.38 and later -spd solves this bug

more info:

http://sourceforge.net/p/sevenzip/bugs/1557/

sc0w:
Fixes #108


monsta:
@raveit65: can you please test it? I don't have 9.38 (Debian and Ubuntu provide only 9.20).

PR #124, opened 2016-02-03
rar: avoid the message "No files to extract"

### LINKED ISSUE #123: engrampa with rar shows the message "No files to extract"
how to reproduce:

1 rar file with one file compressed

I go to "archive" -> "extract..."

unmark "overwrite existing files"

and the button "extract"

If the compressed file already exists, engrampa shows the message "No files to extract"

I have tested with rar 5.30

sc0w:
Fixes #123


raveit65:
Thank you

PR #129, opened 2016-04-30
Give unrar priority over p7zip

svonohr:
At the moment p7zip does not support RAR5 archives, so reading such archives will always fail.

### DIFF
--- a/src/fr-init.c
+++ b/src/fr-init.c
@@ -342,6 +342,7 @@ register_commands (void)
 
 	register_command (FR_TYPE_COMMAND_TAR);
 	register_command (FR_TYPE_COMMAND_CFILE);
+	register_command (FR_TYPE_COMMAND_RAR);
 	register_command (FR_TYPE_COMMAND_7Z);
 	register_command (FR_TYPE_COMMAND_DPKG);
 
@@ -353,7 +354,6 @@ register_commands (void)
 	register_command (FR_TYPE_COMMAND_ISO);
 	register_command (FR_TYPE_COMMAND_JAR);
 	register_command (FR_TYPE_COMMAND_LHA);
-	register_command (FR_TYPE_COMMAND_RAR);
 	register_command (FR_TYPE_COMMAND_RPM);
 	register_command (FR_TYPE_COMMAND_UNSTUFF);
 	register_command (FR_TYPE_COMMAND_ZIP);


raveit65:
The idea looks good for me, on the other side what happen if unrar lakks of support of future RAR versions?
Maybe we need a gesettings key to give unrar or p7zip priority.

svonohr:
Why would `unrar` lack support for future RAR versions? It's the official tool for RAR files. One workaround is to install `unrar` and `rar`. This way the RAR module supports reading and writing and takes priority over the p7zip module.

sc0w:
I think it is unnecessary, engrampa works with rar

on fr-init.c we can read: (line 338)

	/* The order here is important. Commands registered earlier have higher
	 * priority.  However commands that can read and write a file format
	 * have higher priority over commands that can only read the same
	 * format, regardless of the registration order. */

I think rar always have higher priority than p7zip, opening rar files.

Works fine for me

svonohr:
Yes, but only if `rar` and `unrar` both are installed. This is a problem since `rar` is closed source and not necessarily available in all distributions. I would prefer that `unrar` alone takes priority over p7zip.

monsta:
> I think rar always have higher priority than p7zip, opening rar files.

Hmm I don't get it... currently ```FR_TYPE_COMMAND_RAR``` is after ```FR_TYPE_COMMAND_7Z```, so what would make rar have higher priority?

monsta:
See also [upstream report](https://sourceforge.net/p/p7zip/bugs/174/) - they say p7zip 15.09 does support RAR5 format.

sc0w:
p7zip with rar files only works if p7zip-rar is installed (non-free and actually works rar v5)

engrampa doesnt support p7zip-rar

(file-roller supports p7zip-rar)

engrampa only works with rar (non-free) or unarchiver (free and actually works rar v5)

sc0w:
This is the reason p7zip-rar dont works for me on engrampa:

See this bug: #136

p7zip-rar supports rar v5, since 15.14

svonohr:
Since the current p7zip should support RAR5, but there are some problems with the new version I took a look at my original problem again. I installed file-roller and indeed it does prefer unrar over p7zip. However, I don't know why. The order in which commands are registered is unchanged. I will try to find out what they did.


> file-roller supports p7zip-rar

Are you sure? It doesn't work for me.

sc0w:
This is the commit in file-roller:

**Give priority to rar and unrar that supports RAR files better**

https://git.gnome.org/browse/file-roller/commit/?id=160fd2dab3977fa9fdf727cb74e6a23b8fc9d20b

svonohr:
Thanks. However, I think this solution is not very clean and would prefer just reordering the `register_command` calls.

sc0w:
your commit is unnecessary

I have tested and actually,  rar has priority over p7zip-rar

can someone else test it ?

( remember this issue to work latest p7zip-rar version:
https://github.com/mate-desktop/engrampa/issues/136 )

sc0w:
aha

actually rar has priorty over p7zip-rar

but p7zip-rar has priority over unrar

yes, your commit works :)

PR #130, opened 2016-05-02
fix extracting files with single-character names. #125

### LINKED ISSUE #125: engrampa can not extract files with one char in its name (a, b, c, 1, 2, 3 ... )
I tested with drag and drop:

1) If the file is not inside a folder, and exists on the destination dir, engrampa always overwrites without confirmation

2) If the file is inside a folder, engrampa can not extract it

This happens with files with one only char

sc0w:
The patch on file-roller:

https://bugzilla.gnome.org/show_bug.cgi?id=761582

Fixes #125

### DIFF
--- a/src/fr-archive.c
+++ b/src/fr-archive.c
@@ -2911,7 +2911,7 @@ compute_base_path (const char *base_dir,
 		return new_path;
 	}
 
-	if (path_len <= base_dir_len)
+	if (path_len < base_dir_len)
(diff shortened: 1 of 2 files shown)


sc0w:
The commit on file-roller:

https://git.gnome.org/browse/file-roller/commit/?id=64dbaaba67bcaba5af9e570766aebfee7d24db79

PR #140, opened 2016-05-28
require GTK+ >= 3.14, drop GTK+2 code and --with-gtk build option

monsta:
@flexiondotorg @clefebvre @raveit65 @posophe @XRevan86 @willysr @obache @NP-Hardass

Next one... now we move Engrampa to GTK+3 - but with one exception:

Caja extension will be always built with GTK+ version that is used by libcaja-extension. It's autodetected at build time.
It is technically possible because Caja extension doesn't need any code from Engrampa at build time. It only invokes Engrampa binary at runtime.

Other changes are as usual:
- minimum GTK+ version is set to 3.14
- GTK+2 code is dropped
- ```--with-gtk``` build option is dropped as well
- build-dep on GTK+ is changed (that's what you need to do in your distros)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -30,27 +30,9 @@ dnl ==========================================================================
 
 GLIB_REQUIRED=2.32.0
 GIO_REQUIRED=2.25.5
+GTK_REQUIRED=3.14.0
 CAJA_REQUIRED=1.1.0
 JSON_GLIB_REQUIRED=0.14.0
-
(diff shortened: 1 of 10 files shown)


willysr:
+1

raveit65:
Builds and run fine, tested in gtk+3 environment.

monsta:
Dammit... merged the old version of this branch, with two commits instead of a single squashed one :smile:

PR #148, opened 2016-08-05
make the "encrypt header" option insensitive if no password is specified

sc0w:
based on file-roller commit:
https://git.gnome.org/browse/file-roller/commit/?id=6f7f14fa112a7c5b2a64bedeeca9d570ffa90968

and I add some code to work changing the combo box with the compressed file types

(selecting files in caja -> right button -> compress ...)

@monsta @raveit65 please test


sc0w:
I did a better code

I will make new pull request later

PR #149, opened 2016-08-05
make the "encrypt header" option insensitive if no password is specified

sc0w:
based on file-roller commit:
https://git.gnome.org/browse/file-roller/commit/?id=6f7f14fa112a7c5b2a64bedeeca9d570ffa90968

and I add some code to work changing the combo box with the compressed file types

(selecting files in caja -> right button -> compress -> options...)

### DIFF
--- a/src/dlg-batch-add.c
+++ b/src/dlg-batch-add.c
@@ -41,6 +41,9 @@
 #define BAD_CHARS "/\\*"
 #define GET_WIDGET(x) (_gtk_builder_get_widget (data->builder, (x)))
 
+static gboolean has_password = FALSE;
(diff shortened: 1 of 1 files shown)


raveit65:
Works for me with 7zip format.

sc0w:
the checkbox on "encrypt header" works only with supported formats, **.7z** (p7zip) and **.rar** (non-free)

with my commit, with no password, the checkbox is unavailable

sc0w:
merged in https://github.com/mate-desktop/engrampa/commit/dd3529f648ba643924b94857fe16afd37147788d

PR #150, opened 2016-08-09
Folder /data/ui/ renamed to /src/ui/ in .po files

sc0w:
The folder was changed in https://github.com/mate-desktop/engrampa/commit/0d3a5b087f4bb6c118abe34a14ea27b86583f812

### DIFF
--- a/po/af.po
+++ b/po/af.po
@@ -125,100 +125,100 @@ msgstr "Argiefbestuurder"
 msgid "Create and modify an archive"
 msgstr "Skep en wysig 'n argief"
 
-#: ../data/ui/add-options.ui.h:1
+#: ../src/ui/add-options.ui.h:1
 msgid "Load Options"
 msgstr "Laai keuses"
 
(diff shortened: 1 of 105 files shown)


raveit65:
Does creating engrampa.pot and updating the source at transiflex not not the same job?
@monsta ???

sc0w:
I did it with:

`sed -i "s/data\/ui/src\/ui/g" *`

raveit65:
In a copied repo folder do
'./autogen.sh && make && cd po && make engrampa.pot'
Does the created engrampa.pot contain the right paths?
Such a file we submit to transiflex if strings are changed in code.

raveit65:
Problem is that if we change something in po files it will override by next transiflex update.

sc0w:
aha, `engrampa.pot` seems to be fine

I understand :)

raveit65:
Ok, i will upload the new pot file.

Issue #127, opened 2016-03-04
No output file from archiver with parted rar/7z archives over engrampa


treki1:
My system:
Linux Mint 17.3 Rosa 64-bit
Kernel Linux 3.19.0-32-generic x86_64
MATE 1.12.0

Packing over the terminal works fine.
With engrampa (right mouse button), it works only with not parted archives (single file).
If i want to create a parted archive over engrampa, the process works and create a temp folder with the compressed output files.
After processing, the created temp folder will be deleted.
That's so since 17.1. With 17.0 it works correctly.

treki1:
No other peoples have this problem?
In LMDE2 and LM18 is the same :(
If i install engrampa_1.8.0-0 (standard version in LM 17.0), it works.

treki1:
*PUSH*
Still the same bug. Also in Mint Sarah.
If i install the old packages from Qiana
**engrampa-common_1.8.0-0+qiana_all.deb**
and
**engrampa_1.8.0-0+qiana_amd64.deb**
it work.

treki1:
Ok. Wenn Du keine Zeit oder Lust hast, dann gib mir den Code.
Ich werde das in Ordnung bringen.
Es kann ja nicht sein, dass dieses kleine Tool eine ganze Linuxgemeinde in Verlegenheit bringt!

raveit65:
Bitte, https://github.com/mate-desktop/engrampa

treki1:
Was bitte? Was soll der Link? Es muss doch in die Quellen! Es kann nicht sein, dass eine Neuinstallation von LM Mate den fehler integriert hat!

voger:
I have the same problem. Manjaro Linux with Xfce and engrampa 1.16.0

sc0w:
@treki1 in english, please

> That's so since 17.1. With 17.0 it works correctly.

engrampa version that works?

treki1:
The old packages from Qiana (lm17)
engrampa-common_1.8.0-0+qiana_all.deb
and
engrampa_1.8.0-0+qiana_amd64.deb
work with lm17.1 - lm8.1

sc0w:
I can reproduce it on engrampa 1.10.1, and works on 1.8.1

treki1:
I not understand :(
What you do reproduce? And it work? You have rar parts?

sc0w:
I mean: engrampa 1.10.1 have same problem, and it works on 1.8.1

treki1:
[...]

PR #162, opened 2017-02-22
Fix: compress 7z split into volumes

sc0w:
After https://github.com/mate-desktop/engrampa/commit/0b3b7eef61287213308e14639444e1953b52ac04 the compress with volumes doesn't work

This is the fix for 7z files, all the files will be move from the temp folder to the current folder at same time

Fixes partially #127 

and, the fix for rar volumes is pending

### DIFF
--- a/src/fr-process.c
+++ b/src/fr-process.c
@@ -675,14 +675,37 @@ start_current_command (FrProcess *process)
 	char          **argv;
 	int             out_fd, err_fd;
 	int             i = 0;
+	char           *commandline = "";
+	gboolean        fixname = FALSE;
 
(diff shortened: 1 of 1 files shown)


sc0w:
after the PR the compress .tar.7z doesn't work

I rebased, and fixed with these changes:

```
-				if (g_str_has_suffix(argv[2], ".7z")) {
+				if ((g_str_has_suffix(argv[2], ".7z")) && (!g_str_has_suffix(argv[2], ".tar.7z"))) {
```

raveit65:
Confirmed, this fixes the regression with .tar.7z.

PR #164, opened 2017-03-05
Fix: compress rar split into volumes

### LINKED ISSUE #127: No output file from archiver with parted rar/7z archives over engrampa

sc0w:
Fixes #127

### DIFF
--- a/src/fr-process.c
+++ b/src/fr-process.c
@@ -689,10 +689,20 @@ start_current_command (FrProcess *process)
 
 		if (g_str_has_prefix(commandline, "mv")) {
 
-			if ((i==3) && (!g_file_test(g_shell_quote(argv[2]), G_FILE_TEST_EXISTS)) && (!fixname)) {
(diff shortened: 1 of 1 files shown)


raveit65:
As rar is non-free i can't really test it as i don`t have rar.

raveit65:
I found a trial version somewhere.
It works well now, nice work.

raveit65:
Btw. you ask for more reviewer if you like ;-)

sc0w:
thanks for your quick tests

next step: .cbr

Issue #163, opened 2017-03-04
wrong date modified shown in column when using unrar as lib


kn00tcn:
i probably should have mentioned this almost a year ago when i first started using linux

(these are some examples, 'turns into' is engrampa's column, i'll rename the day-month-year it uses so it's easier to see)
`2007-06-04` turns into `2009-11-28`
`2012-09-19` turns into `2025-03-04`
`2013-09-20` turns into `2026-03-06`
`2013-09-21` turns into `2027-03-06`
`2014-07-07` turns into `2013-01-03`
`2014-07-10` turns into `2016-01-04`

it looks like the similar dates show identical year differences, that must be the clue we need

the files are rar version 4 & 5, dont think i've checked version 3, the hours:minutes might be fine

using the 'last output' option shows the correct date, as does commandline unrar, so unrar itself seems to be fine

the extracted files also have the correct date, & it's correct when switching to unarchiver (coincidentally i now found a bug in unarchiver right as i started testing, it takes a min to scan a specific archive, but not other ones, so it would be nice to use the official unrar)

sc0w:
with `rar l thefilename.rar` the date is correct

with `engrampa` and `file-roller` with `rar` is wrong

`engrampa` with `unar` shows the date correctly

this needs some investigation

confirmed

kn00tcn:
ah yes, i did try gnome's archive viewer many months ago, same issue, at the time i assumed it was rar's fault but never checked what it outputs until now

https://github.com/GNOME/gnome-autoar/commit/66cf03f9f1913bc908fea56afbdf984e1a2c18d6 funny

also tried xarchiver now, nothing shows up when opening rars even though it says it's supported

...ok i held off on posting this reply until i searched the code

https://github.com/mate-desktop/engrampa/blob/master/src/fr-command-rar.c line 72, it seems to be doing day-month-year instead of year-month-date

i am not set up to compile & test, can i still make a pull request for someone else to test?

sc0w:
> ah yes, i did try gnome's archive viewer many months ago, same issue, at the time i assumed it was rar's fault but never checked what it outputs until now

the rar commandline works fine, with `rar l filearchive.rar` I can see the real date

> GNOME/gnome-autoar@66cf03f funny

engrampa doesn't support libarchive

> also tried xarchiver now, nothing shows up when opening rars even though it says it's supported

I can't open rar files with xarchiver and rar installed, I tried with xarchiver 0.5.4

[...]

PR #165, opened 2017-03-08
Fix: rar file listing now shows correct date

kn00tcn:
rar before version 5 used DD-MM-YY, but more recent rar versions use YYYY-MM-DD, most people using rar would and should be using recent versions

this fix reorders the string split so year is first and day is last in order to show the correct date when using rar version 5 onward

the newer date format matches the one used by tar, so this fix also adjusts the year line to support four characters instead of two

updated example listing output for both rar versions is included and moved before the time function instead of after

### DIFF
--- a/src/fr-command-rar.c
+++ b/src/fr-command-rar.c
@@ -55,6 +55,43 @@ have_rar (void)
 
 /* -- list -- */
 
+/*
+
+// SAMPLE RAR VERSION 5 LISTING OUTPUT:
+
+RAR 5.40   Copyright (c) 1993-2016 Alexander Roshal   15 Aug 2016
+Trial version             Type RAR -? for help
+
+Archive: test.rar
+Details: RAR 4
+
+ Attributes      Size    Packed Ratio    Date    Time   Checksum  Name
+----------- ---------  -------- ----- ---------- -----  --------  ----
+ -rw-rw-r--      3165      1310  41%  2017-03-07 21:34  888D50B3  loremipsum.txt
+ -rw-rw-r--         0         8   0%  2017-03-07 21:36  00000000  file2.empty
+----------- ---------  -------- ----- ---------- -----  --------  ----
+                 3165      1318  41%                              2
+
+
+
+// SAMPLE RAR VERSION 4 LISTING OUTPUT:
+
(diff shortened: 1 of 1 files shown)


kn00tcn:
initial discussion https://github.com/mate-desktop/engrampa/issues/163

sc0w:
It works, thanks

why the changes on the comments?

I prefer to respect the original comments, this is useful information:

> RAR 5.00 beta 8
> 05-09-**13**

so, the changes on the year is in some point after 5, I will investigate exactly what is the version with this change, and I will investigate to check the version to work with both versions

sc0w:
date behavior changed since rar 5.30

sc0w:
and now, I have a fix for the check version, and works with older/newer versions

kn00tcn:
omg for some reason i couldnt find when the change was, i searched multiple ways, now i see it right there `12. Console RAR "l" and "v" commands display file time in YYYY-MM-DD format.`

so that means november 2015? hmm... awfully recent

do you want me to add one more commented example for v5 to v5.2? the original didnt look right to use a beta, also i made the archive contain 2 files which is relevant to v4's line breaks/indents

sc0w:
remember, the changes in date style is since **5.30** , so, the commit message is wrong

I suggest maintain the current comments of list output, and just to add one more for 5.30

anyway,  if you agree, the easy way, you can forget the comments, with no changes, and I can add it later one more for 5.30 on the check version, ok?

kn00tcn:
i want to make my first commit to a real project ;)

i have updated the title/message/code so mine can be approved followed by you making version check commit

later i'll probably have to tell file-roller to do the same

sc0w:
please, can you revise the commit message and the code again, and use 5.30 instead 5.3, and 5.21 instead 5.2? this will avoid confusions

and, in the commit message, if you like, in the last line, for a better organization, you can add:
**Fixes #numberissue**

and later, I think, this PR will be ready to merge

kn00tcn:
that hashtag is not specific enough, it's more like #dateissue, but any hash would look weird

i specified the versions & made slightly better wording describing the samples

sc0w:
thanks

squashed and merged in https://github.com/mate-desktop/engrampa/commit/e1646ec8dc32fe55fc610c7a17cf6c128a0827cc

PR #166, opened 2017-03-18
extract dialog: always ask whether to overwrite

### LINKED ISSUE #161: Wrong behavior of Skip button in Replace file dialog
Button **Skip** in "Replace file?" dialog extract whole archive instead of skipping one file.

Version: engrampa-**1.16.0**-1.fc25.x86_64
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1422004

1. Prepare ZIP archive containing files A and B and blank directory containing A
2. Open archive using engrampa
3. Drag A to directory - dialog Replace file "A"? will show
5. Press Skip button

**Actual results:**
File A from directory is replaced by file A from archive and file B is extracted too - directory now contain files A and B from archive.

**Expected results:**
In directory should be original file A only. (no file B)

sc0w:
Removed the overwrite option in the extract dialog

based in file-roller commit:

https://git.gnome.org/browse/file-roller/commit/?id=fe71357aa128d029b6ca4a9493b8a408c8aaf017

Fixes #161

### DIFF
--- a/data/org.mate.engrampa.gschema.xml.in
+++ b/data/org.mate.engrampa.gschema.xml.in
@@ -133,10 +133,6 @@
     <child name="batch-add" schema="org.mate.engrampa.dialogs.batch-add"/>
   </schema>
   <schema gettext-domain="@GETTEXT_PACKAGE@" id="org.mate.engrampa.dialogs.extract" path="/org/mate/engrampa/dialogs/extract/">
-    <key name="overwrite" type="b">
-      <default>true</default>
-      <summary>Overwrite existing files</summary>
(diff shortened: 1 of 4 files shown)


raveit65:
Why shouldn't it be allowed to override existing files?
Another more useless question if i want to override a file.

sc0w:
override? do you mean "overwrite" ?

you can do the same with "replace" or "replace all":

dialogengrampa

sc0w:
ups, **segmentation fault** clicking in the "extract" button in the "extract dialog"

I will investigate this

PR #168, opened 2017-03-22
add some style class

raveit65:
This adds 'engrampa-window' at top and the 'frame' style class to the scrolledwindows.
scrolledwindow.frame is used to set a border in most themes.
Adding a specific class at top of the code of an application is always a good idea.
At the moment i need this, as i have to set a border on 'frame > border' for firefox in themes to have a border for the locationbar.
See https://github.com/flexiondotorg/ubuntu-mate-themes/issues/35
In result i can remove the border here with
```
.engrampa-window > grid > paned > box > frame > border {
    border-style: none;
}
```

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -5420,6 +5420,7 @@ fr_window_construct (FrWindow *window)
 {
 	GtkWidget        *menubar;
 	GtkWidget        *toolbar;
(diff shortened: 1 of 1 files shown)


raveit65:
I applied the improvements and rebased the commits.

PR #170, opened 2017-04-21
fix a few deprecations

raveit65:
This affects the add folder dialog and progress window.


raveit65:
Thank you for quick testing :-)

PR #171, opened 2017-04-30
deprecation fixes

raveit65:
Note, i am not really sure if all that ui files are used.....

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -5717,9 +5717,9 @@ fr_window_construct (FrWindow *window)
 	/* * filter entry */
 
 	window->priv->filter_entry = GTK_WIDGET (gtk_entry_new ());
-	gtk_entry_set_icon_from_stock (GTK_ENTRY (window->priv->filter_entry),
-				       GTK_ENTRY_ICON_SECONDARY,
-				       GTK_STOCK_CLEAR);
+	gtk_entry_set_icon_from_icon_name (GTK_ENTRY (window->priv->filter_entry),
+				           GTK_ENTRY_ICON_SECONDARY,
+				           "edit-clear");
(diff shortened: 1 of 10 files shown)


sc0w:
@raveit65 

`fr-window.c` Did you forget to delete the commented lines after your tests?

`password.ui` (edit->password) seems to be better with this PR

without the PR

passwordbefore

with the PR

passwordlater

`delete.ui` (pressing key DEL in a folder/file) I see the same than @lukefromdc 

without the PR

deletebefore

with the PR

deleteafter

raveit65:
> fr-window.c Did you forget to delete the commented lines after your tests?

Opps, the commented lines are deleted now.

> password.ui (edit->password) seems to be better with this PR

Yes, i added "max_width_chars">70 for the "Note......." label.
I think 70 chars is a good value, i will use that for other applications too.

raveit65:
delete.ui;
For some reasons the GtkEntry doesn't expand over the column width here after porting GtkTable to GtkGrid. The result looks like that. (proposal 1)
engrampa-delete-ui-after-grid
In previous comment i packed the GtkLabel Files and the GtkEntry in a horizontal GtkBox. which breaks the Checkboxes.
It's possible to set a width request for the GtkEntry (proposal 2)
engrampa_delete-ui-entry-large
I can also revert the whole commit and we use deprecated Gtkable here (proposal 3), sorry i've no idea to do that better.
Thoughts?

raveit65:
I found a fix for delete.ui, now it looks like this.
engrampa-delete-ui-fixed
PR is rebased, please test again,

lukefromdc:
This now works, the radion button problem in the delete dialog is fixed

PR #173, opened 2017-05-01
Fix: Wrong behavior of Skip button in Replace file dialog

### LINKED ISSUE #161: Wrong behavior of Skip button in Replace file dialog

sc0w:
based in file-roller commit:
https://git.gnome.org/browse/file-roller/commit/?id=fe71357aa128d029b6ca4a9493b8a408c8aaf017

Fixes #161 

I take the needed in that file-roller commit, and I keep the overwrite option in the extract dialog

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -3525,6 +3525,24 @@ fr_window_get_file_list_pattern (FrWindow    *window,
 }
 
 
+static GList *
+fr_window_get_file_list (FrWindow *window)
+{
+	GList *list;
+	int    i;
+
+	g_return_val_if_fail (window != NULL, NULL);
+
+	list = NULL;
+	for (i = 0; i < window->archive->command->files->len; i++) {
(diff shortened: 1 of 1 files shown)


sc0w:
merged in https://github.com/mate-desktop/engrampa/commit/dff9db3296b4d0e7711271f429a70c7340a68e85

PR #174, opened 2017-05-04
hide folders in "View All Files"

### LINKED ISSUE #172: "View All Files" view-trying to delete selected folder or folders deletes everything in archive
If and only if the "view all files" option for Engrampa is selected, the delete dialog gives the option to delete selected files, which works fine for files but if a directory is selected, both all directories and all files in the toplevel directory are deleted. What remains is an empty archive.

sc0w:
Fixes #172

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -1117,9 +1117,8 @@ compute_file_list_name (FrWindow   *window,
 		return FALSE;
 
 	if (window->priv->list_mode == FR_WINDOW_LIST_MODE_FLAT) {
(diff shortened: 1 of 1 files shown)


raveit65:
@monsta @sc0w @lukefromdc 
For me this belongs in a 1.18.1 point release as it is a bug fix, on the other side it change a bit the look&feel.
Thoughts?

sc0w:
@raveit65 

Yes, it changes the look, but we have a reason, so I agree

PR #176, opened 2017-05-07
Fix: Browsing history not correct

### LINKED ISSUE #167: Browsing history not correct
Browsing through an archive and then pressing the 'back' button back to home again greys out the back button, as it should, but then if I browse to something else, say another folder in the archive, and *then* press back, engrampa goes through not just the one folder browsed, but all the folders that had been previously browsed before hitting the back button the first time.
Hope that all makes sense.  It's not a big issue, just kind of annoying.

sc0w:
Fixes #167

based in file-roller commit:
https://git.gnome.org/browse/file-roller/commit/?id=bd8807beb9a026be2441aa492815b65a6fc94429

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -869,23 +869,38 @@ static void
 fr_window_history_add (FrWindow   *window,
 		       const char *path)
 {
-	if ((window->priv->history != NULL) && (window->priv->history_current != NULL)) {
-		if (strcmp (window->priv->history_current->data, path) == 0)
-			return;
+	if ((window->priv->history_current == NULL) || (g_strcmp0 (path, window->priv->history_current->data) != 0)) {
+		GList *scan;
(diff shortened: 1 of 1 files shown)


raveit65:
What is current behaviour?
What is expected behaviour?
History seems OK if i press the forward/back buttons and using a sidebar.

sc0w:
This is the issue:

[issue167.zip](https://github.com/mate-desktop/engrampa/files/981456/issue167.zip)

sc0w:
easy to reproduce, with this sample zip file:

[isuue167.zip](https://github.com/mate-desktop/engrampa/files/981833/isuue167.zip)

1. double click in **folder**

2. click in **Back**

3. double click in **folder**

4. click in **Back**

and see the back button:

without this PR:

backbad


with this PR (the expected):

backgood

PR #178, opened 2017-05-20
Fix: create zip files in "maximum" compression level

sc0w:
based in file-roller commit:
https://git.gnome.org/browse/file-roller/commit/?id=f9b8a010e7fe5a9bec158dfc5bd8ef1e11bf3842

Fixes:
https://bugs.launchpad.net/ubuntu/+source/engrampa/+bug/1580860

### DIFF
--- a/src/fr-command-7z.c
+++ b/src/fr-command-7z.c
@@ -362,7 +362,11 @@ fr_command_7z_add (FrCommand     *comm,
 		break;
 	case FR_COMPRESSION_MAXIMUM:
 		fr_process_add_arg (comm->process, "-mx=9");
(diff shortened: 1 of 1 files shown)


monsta:
I wonder what that code does in 7z command file instead of zip command one...

sc0w:
> I wonder what that code does in 7z command file instead of zip command one...

yes, 7z does the job

> Side note: I was pretty sure we could set the compression level in the dialog, not just in dconf-editor...

No, but it is a good feature in the future, using `caja` with `right click -> compess` we have the code in `dlg-batch-add.c` and `batch-add-files.ui`

PR #179, opened 2017-05-20
Restore Polish translation credits from GNOME

piotrdrag:
This commit restores proper credits and copyrights for the Polish translation that were somehow lost during forking, using historical information from git.gnome.org.


raveit65:
merged https://github.com/mate-desktop/engrampa/commit/5fec8a8951c93a900e952217b8f58cf12943f0c4
Thank you

Issue #185, opened 2017-08-12
delete files in a encrypted archive


mirkander:
#### Expected behaviour
Delete or rename files inside encrypted archive (rar o 7z) created selecting "encrypt the file list too" option.

#### Actual behaviour
Engrampa doesn't do it.

#### MATE general version
1.18

#### Package version
1.18.2-1

#### Linux Distribution
Linux mint mate 18.2

monsta:
Fix for 7z is released in version 1.18.3. Fix for rar will be done later. If you know any other types of archives where this issue should be fixed, please post them here.

sc0w:
rename files in 7z without the list encrypted is fixed here: https://github.com/mate-desktop/engrampa/commit/516bd1a9d6003711b0ba42e3dcd233997f6cf777

rar is pending

monsta:
Cool, I've picked the latest commit for 1.18 branch too.

zhuyaliang:
@mirkander Please test if the issue still exists

PR #188, opened 2017-10-11
7z: Fix: delete/rename files/folders with the list encrypted

sc0w:
based in file-roller commit:
https://git.gnome.org/browse/file-roller/commit/?id=0bb26ee4223cdcbca9aa0816c81069b74f15c2f7

Fixes partially https://github.com/mate-desktop/engrampa/issues/185

### DIFF
--- a/src/fr-command-7z.c
+++ b/src/fr-command-7z.c
@@ -422,6 +422,8 @@ fr_command_7z_delete (FrCommand  *comm,
 			if (g_str_has_prefix (scan->data, "@"))
 				fr_process_add_arg_concat (comm->process, "-i!", scan->data, NULL);
 
+	add_password_arg (comm, FR_COMMAND (comm)->password, FALSE);
+
 	fr_process_add_arg (comm->process, "--");
 	fr_process_add_arg (comm->process, comm->filename);
 


sc0w:
@lukefromdc thanks for testing

merged: https://github.com/mate-desktop/engrampa/commit/b4cb14f4340f26ec448f78841dbb84d605fde0eb

> Another PR will then be needed to make Engrampa call for the password when a file is deleted or renamed when the file list is NOT encrypted

Yes, I can't rename/delete files if the list isn't encrypted

lukefromdc:
You can rename/delete if the list isn't encrypted, by opening one file first to force the passphrase dialog. This is not discoverable by a typical user however

monsta:
So that part is needed to be fixed to get https://github.com/mate-desktop/engrampa/issues/185 closed, right?

sc0w:
@monsta the issue only mentioned the encrypted files, so, the fix for **rar** files is pending to close the issue, I think

monsta:
Ah, I forgot about rar. But I actually meant the issue you discussed in the last comments.

sc0w:
> Ah, I forgot about rar. But I actually meant the issue you discussed in the last comments.

It is a pending task, but it is not needed to fix the issue, because only was mentioned encrypted files, I think

monsta:
It wasn't mentioned - but would be nice to fix someday 🙂

sc0w:
@monsta yes, of course

PR #192, opened 2018-01-28
require GTK+ 3.22 and GLib 2.50

monsta:


wdlkmpx:
I think there is a better way to deal with deprecated stuff, and it's called version abstraction... or something like that (can't remember), here is an example:
https://lists.geany.org/pipermail/github-comments/2016-January/002026.html

File-roller/engrampa is (probably) the best gtk archiver app for linux, at least the most friendly/windows-like. So people outside gnome and derivatives may want to compile that specific app in some other DE's and/or older systems. Following gnome's steps is usually the wrong approach. Just saying.

I was able to apply all the important fixes to an old engrampa version which is a commit before gtk2  was dropped, the thing compiles ok in old and new systems and works like a charm under any DE:
https://github.com/wdlkmpx/engrampa/commits/master

lukefromdc:
Well, that's one of the nice things about open source and about git: if you need to do something different from what the main project is doing, you can. MATE itself and Cinnamon too were born this way.

As for why we've settled on GTK 3.22 and GLib 2.50 as the only supported versions, it's because these are what current distros likely to ship MATE 1.20 use and our team is small. We don't have the time as a team to test against five versions of GTK3 plus GTK2, and as a result the old stuff would go untested, get buggy, at some point might stop working and nobody would know when, where, or why it broke.   By comparison, if you are actively maintaining a GTK2 version of Engrampa you will know right away if something breaks it, because that's what you are testing with.  From time to time you might see a commit that does something useful or fixes a bug but uses code only compatable with GTK 3.22. As the maintainer of the forked version, you can drop that commit, port it to GTK2, or write new code from scratch to do the same job, whichever best suits your needs.

At some point in the future distros might drop GTK2 altogether, and GTK 3.22 is the LTS version of GTK3 as the GNOME devs have now moved on to developing GTK4, expected out this fall. I don't expect us to attempt to port to that anytime soon, as GTK4 removes a lot of code needed for almost all previously written custom widgets. It will be a very long time before GTK3 goes unsupported and at that time everyone will know what subsequent version of GTK is the next LTS version. GNOME themselves branched 4.0 because of the GSK (scene graph) commits which would have broken Libreoffice and Firefox which had just ported to GTK3. They now want to increment major versions more frequently so different versions can be installed alongside oneanother.  Time will tell what their next LTS version will be,

PR #199, opened 2018-04-04
fix crash on exit after using Open With dialog

monsta:
Steps to reproduce:

- open some archive
- right-click on some file and choose Open With
- either open file with some app, or just close the dialog (cancel)
- exit Engrampa

Without the fix, it will crash on exit with double-free error. The patch removes excessive freeing of GList which contains the list of files.

### DIFF
--- a/src/dlg-open-with.c
+++ b/src/dlg-open-with.c
@@ -39,14 +39,6 @@ typedef struct {
 
 
 static void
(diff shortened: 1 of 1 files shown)


lukefromdc:
Error confirmed giving this message:

free(): invalid pointer
Aborted

PR #209, opened 2018-07-12
UI: on the properties dialog, focus the Close button by default

kn00tcn:
the Close button, not the Help button, should be the one set to default so that you can press Enter or Space without unintentionally opening the help manual


kn00tcn:
i guess github wants to cut my original title after 70chars

raveit65:
template ignored....

raveit65:
oops

raveit65:
Thanks, this works fine.

PR #212, opened 2018-08-01
Add support for OpenDocument formats

### LINKED ISSUE #193: engrampa does'nt open odt files
#### engrampa does'nt open those files (while trying to open, engrampa reports:...engrampa does not support odt-type of archives).

file-roller can open these files, because odt files are sort of zip-archives.

vkareh:
Fixes #193


cranes-bill:
There is much more other formats which in fact are archives, but in general they are not intended to be opened as archive. Do you plan to put all of them here? That is wrong.
If somebody wants to open such a file, it's enough just to rename it... 
**(IMO)**

lukefromdc:
Personally I find such archives as .apk files (which I consider equivalent to a Debian file in a way) having to be renamed to open them to be a nuisance. This I am in favor of the graphical archive managers being able to open anything that can be unpacked with tar from the command line.

In the .apk case, opening them is a necessary step to using jarsigner to verify that a sideloaded (non-Google Play) apk file has not been tampered with in an MITM attack. Having to rename the file is a nuisance extra step.

PR #216, opened 2018-09-04
%'d -> %d

houbaron:
In [engrampa.pot](https://github.com/mate-desktop/engrampa/blob/master/engrampa.pot#L980), The "msgid_plural" was defined as "%'d files remaining". But we all know it should be "%d".

It's Chinese, forgive me.

More than that, all translation refer the engrampa.pot. So all of them are wrong.

### DIFF
--- a/engrampa.pot
+++ b/engrampa.pot
@@ -977,7 +977,7 @@ msgstr ""
 #: ../src/fr-window.c:2708
 #, c-format
 msgid "%d file remaining"
-msgid_plural "%'d files remaining"
+msgid_plural "%d files remaining"
 msgstr[0] ""
 msgstr[1] ""
 
--- a/po/af.po
+++ b/po/af.po
@@ -999,7 +999,7 @@ msgstr ""
 #: ../src/fr-window.c:2708
 #, c-format
 msgid "%d file remaining"
-msgid_plural "%'d files remaining"
+msgid_plural "%d files remaining"
 msgstr[0] ""
 msgstr[1] ""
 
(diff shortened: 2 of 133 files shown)


vkareh:
@houbaron - Translations are handled in Transifex. The only one that seems to belong here is the change in `fr-window.c`: https://github.com/mate-desktop/engrampa/pull/216/files#diff-323a9cb3946850419e43e05b60100e9dR2709

Everything else can be removed from this pull request. @raveit65 - do you agree?

raveit65:
Yes, i agree that translation commit needs to removed from PR by author.
But i do not agree with Opening PRs without an description.
I don't want to read a book, but...............

houbaron:
@lukefromdc @vkareh @raveit65 Please watch this: . It's Chinese, forgive me.
All numbers in progress bar are replaced by "%'d".

lukefromdc:
This does not occur in English. I know almost nothing about translation handling but I think it does not occur here because no translation is involved. Report is that all other languages get this issue so problem must be in a file used by all translations but only in translation.

raveit65:
We generate engrampa.pot in /po dir with that script
https://github.com/mate-desktop/engrampa/blob/master/makepot

raveit65:
I guess the typo needs to be fixed here https://github.com/mate-desktop/engrampa/blob/master/src/fr-window.c#L2709
After that ./makepot should generate engrampa.pot without this error.

raveit65:
Done!

raveit65:
Thanks for reporting this issue :)

PR #220, opened 2018-09-30
Fix url of ulinks to point to mate-user-guide

rbuj:


raveit65:
Thanks for pointing that out, i cherry-picked that to 1.20.

PR #221, opened 2018-10-03
Manual: Update file format descriptions using shared-mime-info

rbuj:
REF: https://cgit.freedesktop.org/xdg/shared-mime-info/tree/freedesktop.org.xml.in


raveit65:
Thank you

PR #223, opened 2018-10-04
Use make functions for HELP_LINGUAS

rbuj:


raveit65:
@lukefromdc 
We pulling translations from transifex sever with `tx pull -a -f` before we make a new release.
Doing this the first time after we added engrampa help to transifex, will pull in a lot of **new** languages to 
`engrampa/help/` dir.
Without the PR you need to edit this line in the `engrampa/help/Makefile.am` to support the new languages.
https://github.com/mate-desktop/engrampa/blob/master/help/Makefile.am#L12
`HELP_LINGUAS = bg ca cs de el en_GB es fi fr it ja oc nl ru sv vi uk zh_CN`
So the PR does simply read all existing languages/folders in this dir, and add the result to the make file.
In result all new languages and po files are in a new generated tarball with
`./autogen.sh && make && make distcheck`

lukefromdc:
Thanks for the reply, in that case I favor all of these PR's. Since I'm not the one doing the transifex work I can't test that but these are all fine with me

PR #225, opened 2018-10-27
Added test integrity for the cfile compressors

stokito:
Added test integrity for the cfile compressors: gzip, bzip2, etc. 
But since most of them shows the message with file status to STDERR instead of STDOUT whe should show both in Test result window.

To validate:
Open *.gz file (or *.bz2, *.lzip...), then File/Test integrity. You should see output of test command.
The only one archive which doesn't support test is rzip.

Also in #225 contains support of test brotli but this needs for #224 to be merged first.


stokito:
I didn't understood what do you mean "test a .xz file was greyed out ". I know only one place to trigger the integrity: main menu/Archive/Test integrity
If the menu item is grayed then maybe that's because `xz` is not installed in your system.
Then try to install it. You can test it on `brotli` for example too.

lukefromdc:
I have .xz support on my system and use it all the time. The engrampa menubar's archive>test integrity menu item is the one I had greyed out. Same thing happened with xz in #226 for some reason, even though I can both pack and unpack .xz archives with no problem in Engrampa on my system.

stokito:
Hi @lukefromdc 
The `Test Integrity` menu item was grayed because this wasn't supported until this PR. If you still see grayed item then probably you didn't `sudo make install` or try to remove `engrampa` and `engrampa-common` from Ubuntu. The only one unsupported integrity is rzip, but it will show the menu item and on click show empty result dialog anyway.
Meanwhile I see that the `xz` package name referred in sources is not correct and created #227 to fix that.
So, could you accepts the PR and then related  #226?

src/fr-command-cfile.c
```diff
@@ -594,6 +627,7 @@ fr_command_cfile_class_init (FrCommandCFileClass *class)
 	afc->add              = fr_command_cfile_add;
 	afc->delete           = fr_command_cfile_delete;
 	afc->extract          = fr_command_cfile_extract;
+    afc->test             = fr_command_cfile_test;
```
sc0w: the indentation is wrong here and before

lukefromdc:
I am not running Ubuntu but rather Debian Unstable without any of Debian's default MATE packages. I have all of MATE locally compiled

lukefromdc:
This may depend on https://github.com/mate-desktop/engrampa/pull/227 in order to work for .xz files.

src/fr-command-cfile.c
stokito: indentation was fixed in last commit of  #226

lukefromdc:
I just got this build error after pulling the branch:

```
  CCLD     engrampa
/usr/bin/ld: fr-command-cfile.o: in function `fr_command_cfile_add':
/home/luke/Desktop/engrampa/src/fr-command-cfile.c:238: undefined reference to `r_process_set_working_dir'
collect2: error: ld returned 1 exit status
/usr/bin/ld: fr-command-cfile.o: in function `fr_command_cfile_add':
/home/luke/Desktop/engrampa/src/fr-command-cfile.c:238: undefined reference to `r_process_set_working_dir'
collect2: error: ld returned 1 exit status

```

stokito:
fixed, pls merge

lukefromdc:
@raveit65 , @monsta , can someone else have a look at this?

stokito:
tar.bzip2 and .zip are not affected by the change. And yes, brotli will show empty output until https://github.com/google/brotli/issues/729 fixed.

src/fr-command-cfile.c
```diff
@@ -469,6 +469,42 @@ fr_command_cfile_extract (FrCommand  *comm,
...
+	else if (is_mime_type (comm->mime_type, "application/x-lzma")) {
+		compress_cmd = "lzma";
+	}
+	else if (is_mime_type (comm->mime_type, "application/x-xz")) {
+		compress_cmd = "xz";
+	}
+	else if (is_mime_type (comm->mime_type, "application/x-lzop")) {
+		compress_cmd = "lzop";
+	} else { // i.e. if (is_mime_type (comm->mime_type, "application/x-rzip"))
+		g_warning ("Test integrity in unsupported for %s\n", comm->mime_type);
```
sc0w: why not greyed out in the gui instead the warning?

stokito: @sc0w because the `fr-command-cfile.c` contains support  for multiple compressors and to make grayed item in GUI we need to set `propTest = FALSE` in `fr_command_cfile_init` function but this affects all compressors.
But we can simply split call compressors from `fr-command-cfile.c` to separate files like `fr-command-gzip.c`, `fr-command-bzip2.c`, `fr-command-rzip.c` etc.
And IMHO it's better to do that even if then we'll have some code duplication.

stokito: Even if we decide to split it to a separate files this will be done in scope of another PR. But until we done that let's live with the message in logs.

stokito:
This branch is ready to be merged. Is anything needs to be done? Thanks

lukefromdc:
No idea. @raveit65 or @monsta might know more

PR #226, opened 2018-10-27
Brotli cfile test

stokito:
depends on #224 and #225


stokito:
yes, that's because brotli really prints nothing to output and I created an issue about this
https://github.com/google/brotli/issues/729

stokito:
I merged this patch into #225

PR #227, opened 2018-11-01
Fixed xz package name

stokito:
The correct name for `xz` package is `xz-utils`:
https://packages.debian.org/search?keywords=xz-utils&searchon=names&suite=stable&section=all

### DIFF
--- a/data/packages.match
+++ b/data/packages.match
@@ -27,6 +27,7 @@ unrar=
 unstaff=
 unzip=
 xz=
+xz-utils=
 zip=
 zoo=
 
--- a/src/fr-command-cfile.c
+++ b/src/fr-command-cfile.c
@@ -569,7 +569,7 @@ fr_command_cfile_get_packages (FrCommand  *comm,
 	else if (is_mime_type (mime_type, "application/x-lzma"))
 		return PACKAGES ("lzma");
 	else if (is_mime_type (mime_type, "application/x-xz"))
-		return PACKAGES ("xz");
+		return PACKAGES ("xz,xz-utils");
 	else if (is_mime_type (mime_type, "application/x-lzop"))
 		return PACKAGES ("lzop");
 	else if (is_mime_type (mime_type, "application/x-rzip"))
--- a/src/fr-command-tar.c
+++ b/src/fr-command-tar.c
@@ -1118,7 +1118,7 @@ fr_command_tar_get_packages (FrCommand  *comm,
 	else if (is_mime_type (mime_type, "application/x-lzma-compressed-tar"))
 		return PACKAGES ("tar,lzma");
 	else if (is_mime_type (mime_type, "application/x-xz-compressed-tar"))
-		return PACKAGES ("tar,xz");
+		return PACKAGES ("tar,xz,xz-utils");
 	else if (is_mime_type (mime_type, "application/x-lzop-compressed-tar"))
 		return PACKAGES ("tar,lzop");
 	else if (is_mime_type (mime_type, "application/x-7z-compressed-tar"))


stokito:
It really doesn't affect anything and I tested and xz files still working as previously. Safe to merge

stokito:
The packages.match file is used when you trying to open an archive which compressor is not installed in system so engrampa shows you a dialog window `dlg-package-installer.c` and asks you to install the compressor. But in fact xz-utils is present in almost all distros by default. For example it's used in .dep files. This may be changed in feature because, for example Ubuntu has plans to replace xz with zstd for packages.

vkareh:
@stokito - if `xz-utils` is _not_ installed in your system, what does engrampa do when trying to open an xz package? Currently if I remove xz-utils, it tries to pull almost the entire system with it, so I cannot really test this properly...

stokito:
Yes, it's not possible to remove xz-utils atm and the only distro without xz is OpenWrt but to see the behavior you can try to open, for example, any [*.rar` file](https://github.com/KDE/ark/raw/master/autotests/plugins/cliunarchiverplugin/data/empty_folders.rar) and engrampa will ask you to install `rar` package.

So there is no issue today but it may become a problem in feature. I guess that you can even remove the line without new problems. But this is just a semantically error.

vkareh:
Okay, I created a throw-away VM and compiled engrampa with packagekit support. I removed `xz-utils` and tried to open an `.xz` file. On `master` I get an error that it cannot find `xz` package. On #227 I get a proper search for `xz-utils` (couldn't actually install it since removing `xz-utils` really messed up something in the VM, but it now searches for the correct package). This change seems correct, based on that.

@stokito - I noticed that in `src/fr-command-tar.c`, line 1121, you also need to update that one to `xz-utils` to account for `application/x-xz-compressed-tar` files. If you can make that change, I think this PR should be good to go.

raveit65:
i agree with @sc0w not to change that without any reason.

stokito:
@raveit65 this is an error, so why not fix it? There is at least two reasons: 
1. it's semantically incorrect so can misleading someone who tries to understand the code: should I put here a program name or a package name? This is a real issue which I faced when tried to figure out this when added `brotli` support.
2. In feature the `xs-util` package may be not present by default in system. Or Engrampa can be ported to other systems like MacOS or Windows where xz-utils isn't present.

sc0w:
Why the name of debian package?

In fedora its name is **xz** (https://apps.fedoraproject.org/packages/xz)

vkareh:
@sc0w - that's a good point, I didn't realize that. I wonder if it can be solved by changing to
```c
return PACKAGES ("xz,xz-utils");
```
That way it would search for both. This seems like a slippery slope, though, as we'll never know all possible naming conventions for `xz`...

stokito:
Ok, so the last change I guess should satisfy everyone

raveit65:
Did anyone check the other 100 linux distros ?    :)

stokito:
Ok, then if it turned out that xz actually was a correct package name at least for Fedora and Arch then let's just close the PR.

vkareh:
Yes, okay, I've been looking at this all morning. Here's what I've found: the `packages.match` file seems to be meant for distros to override which package to look for. So, for Debian-based distros, their install scripts should end up having:
```ini
...
xunzip=
xz=xz-utils
zip=
...
```
The issue is that the current implementation of name lookup in engrampa is broken. It needs these two commits for the name lookup to work:
* https://gitlab.gnome.org/GNOME/file-roller/commit/ed16ec0fdbe73e846869f12664b3ae899624d888
* https://gitlab.gnome.org/GNOME/file-roller/commit/95aafc3c70f75e4a3b70619f6d0471c9240733a1

vkareh:
The other thing is that most front-ends for PackageKit just do a search, so `xz` should also return `xz-utils` (along with possibly other results), so there shouldn't be an issue with leaving it as `xz` (it's been like this in file-roller for 8 years and distros haven't complained)

PR #228, opened 2018-11-01
Added apt install command to install build deps

stokito:

### DIFF
--- a/README
+++ b/README
@@ -71,6 +71,11 @@
     * gtk+ >= 3.14.0
     * libcaja-extension >= 1.1.0 (optional)
 
+  You can install them with the following command:
+  sudo apt install -y mate-common yelp-tools gtk+-3.0 libcaja-extension-dev
+
+  See the HACKING and README-DISTRIBUTIONS for more details about development.
+
 * Install
 
   ./autogen.sh


lukefromdc:
This is specific only to Debian, as apt is the Debian higher level package management system used only by Debian-based distros and no others

stokito:
Ok, but it would be better to have that command at least for Debian\Ubuntu and add for other distros latter.

raveit65:
Do we really print package-manager specific commands here?
For half the linux world this information is useless.
Simply listing the build dependencies should be enough.
I expect from everyone who use linux to know the simplest package-manager commands, before he/she starts to compile a package ;)

stokito:
@raveit65 In fact no. For example I'm Java developer with only basic knowledge Linux development. About 7 years ago I developed a small plugin for Pidgin and for me it's really no so clear what should I do to compile the project. But I wanted is to make a simple contribution just for fun.
And I really spent some time to realize what should I do to compile the engrampa.
When I runned autogen.sh it said me that "mate-common is missing".
What is that? Why it's missing? 
Well, I just hopefully tried `apt install mate-common` and then autogen said me that yelp-tools is missing.
Great! So I need just to install missing dependencies.
But then it asked me to install libcaja-extension but he package was already installed.
And it turned out that I need to install `libcaja-extension-dev` package.

My point is that if we can simplify life for a newcomers then we should do that.
If you think that a command for Ubuntu/Debian is not enough then let's just add another command for rpm and this will cover already 90% of users.

raveit65:
Not sure why a list of dependencies isn't enough for you?
`mate-common yelp-tools gtk+-3.0 libcaja-extension-dev`
I mean adding `apt install -y` doesn't work for a lot of other distros.
An why we should only give this info for debian based distros?
And when we add commands for rpm based distros it doesn't help for ArchLinux, gentoo, bsd based distros and others.
I don't think it is useful that we add commands for whole the Linux world, because we will always forgot one distro.

But if you think debian/ubuntu users aren't smart enough to use their package manager go ahead ;)
I won't block this but the benefit is doubtful to me.

lukefromdc:
I think we could end up on a treadmill trying to provide distro-specific instructions. There are just too many different distros for this, and online searches will usually return the needed results

PR #232, opened 2018-11-08
Fix implementation and use of the alternative package name lookup

vkareh:
Any `real_name` which was found was not getting used before.

Also, when the lookup is unsuccessful, this means there is no
_alternative_ package name.  Thus we should use the hard-coded package
name as the real_name for our PackageKit installation request.

[bug #603757]
origin commit: https://gitlab.gnome.org/GNOME/file-roller/commit/95aafc3c

### DIFF
--- a/src/dlg-package-installer.c
+++ b/src/dlg-package-installer.c
@@ -139,10 +139,12 @@ get_packages_real_names (char **names)
 		real_name = g_key_file_get_string (key_file, "Package Matches", names[i], NULL);
 		if (real_name != NULL)
 			real_name = g_strstrip (real_name);
-		if ((real_name == NULL) || (strncmp (real_name, "", 1) == 0))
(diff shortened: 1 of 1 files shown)


vkareh:
This should supersede #227, and distros that use `xz-utils` (Debian et al) should technically update their local copy of the `packages.match` file to reflect this

stokito:
BTW there is a lot of such things that can be backported to the File Roller and verse versa. As I understood the only real difference between FR and Engrampa is GUI so to avoid code duplication it would be great to extract shared library like libarchive. XFace also have it's own archive manage XArchiver and it even better in some aspects than FR/Engrampa.
Maybe we should discuss this approach instead of cherry picking some commits from FR?

raveit65:
@sc0w 
ready to go?

PR #235, opened 2018-11-25
add start button

zhuyaliang:
I added a status button to pause or start


sc0w:
what does exactly the new file `/src/tags` in this PR ?

zhuyaliang:
@sc0w @lukefromdc
Sorry, I uploaded the wrong code, this is the test code,/src/tags  is a useless file. I will upload the correct code again.

zhuyaliang:
@sc0w @lukefromdc This is the correct code. Please test it.

zhuyaliang:
@lukefromdc I made a change and clicked the Cancel button can kill the suspended process.

lukefromdc:
OK, that last change fixes the freeze, this seems good to go so far as behavior is concerned. 

Not 100% sure about the memory management, but the only *pointer variable I found is part of  _FrWindowPrivateData and seems to be handled the same way as everything else in it. Probably good to go on memory management as well

raveit65:
@zhuyaliang 
Are you familar with the `git rebase -i <last-commit-id-before-PR> ?
With that command you can edit/squash/reword/re-order existing commits in your PR.
Adding and removing unneeded `/src/tags` doesn't look very nice.
I think you will love this useful and powerful `rebase' command.

src/fr-window.c
```diff
@@ -2283,6 +2284,31 @@ fr_window_view_extraction_destination_folder (FrWindow *window)
 	open_folder (GTK_WINDOW (window), fr_archive_get_last_extraction_destination (window->archive));
 }
 
+static void change_button_label(GtkWidget *button)
+{
+	gchar *state;
+	state = gtk_button_get_label(GTK_BUTTON(button));
```
raveit65: I got a build warning here.
```
fr-window.c: In function 'change_button_label':
fr-window.c:2298:8: warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  state = gtk_button_get_label(GTK_BUTTON(button));
        ^
```

raveit65:
It is working very nicely here.

src/fr-window.c
zhuyaliang: @raveit65 I've modified it, using const modifier，Thank you for your help. I'm learning how to use git-rebase.

src/fr-process.c
```diff
@@ -1026,8 +1027,49 @@ fr_process_start (FrProcess *process)
...
+static int
+fr_close_suspend_process(FrProcess *process)
+{
+    int ret = -1;
+    g_return_val_if_fail(process != NULL, ret);
+
+    if (process->priv->suspend)
+    {
+        if (process->priv->command_pid > 0)
+            ret = killpg (process->priv->command_pid,SIGKILL);
```
monsta: Maybe `SIGTERM`? It gives process a chance to exit cleanly.

Also I'm not sure, is it allowed to send such signals to a process which is stopped by `SIGSTOP`?

PR #236, opened 2018-11-28
Adding pause and start functions

zhuyaliang:
@raveit65  @monsta  @lukefromdc @sc0w 
For some special reasons, I messed up the contents of the warehouse, so I could only launch a new pull requests。The code has not changed.。


monsta:
Ok, so the stopped process should be force-killed by `SIGKILL`. But could we lose some data because of that? `SIGCONT` followed by `SIGTERM` would still be cleaner.

zhuyaliang:
@monsta
 Using 'SIGKILL' signal will not cause data loss, it will only force to kill the specified process, If you're worried ， I can use 'SIGTERM' signal and then send 'SIGCONT', which can also end the process.

zhuyaliang:
@monsta
 
``ret = killpg (process->priv->command_pid,SIGTERM);``
``ret = killpg (process->priv->command_pid,SIGCONT);``

zhuyaliang:
@sc0w @monsta 
Is there anything else that needs to be changed? Can accept my application for merger？

monsta:
@sc0w: do you mean renaming the button to Pause/Resume or something? I think it would sound better.

sc0w:
@monsta yes, the names can be discussed, with one letter for using the underline property, and the images "media-playback-pause" and "media-playback-start", and... I think the progress bar and the label "please wait" can be hidden if the button "suspend" is pressed.

PR #241, opened 2018-12-09
fr-window: avoid 'gtk_dialog_add_button' with stock ids

sc0w:


raveit65:
Which dialog is it?

sc0w:
The progress dialog, and I noticed one issue...



The pause button must be hidden here, I will fix it in other PR...

Issue #218, opened 2018-09-07
Filename display issues in flat view for DEBIAN directory contents of .deb files


gm10:
#### Expected behaviour
Proper filename display in flat view for DEBIAN directory contents of .deb files.

#### Actual behaviour
Garbled file names for contents of the DEBIAN directory of any .deb file in flat view only:
![](https://i.imgur.com/37GV03T.png)
In folder view all is correct:
![](https://i.imgur.com/4preqkj.png)
I have not noticed this behaviour with other archive types, including .tar.xz files.

#### Steps to reproduce the behaviour
Open any .deb file, switch to flat view (Ctrl+1)

#### MATE general version
1.20

#### Package version
Tested with 1.20.0-1 from Ubuntu's repo as well as the 1.21 release sources from here. 

#### Linux Distribution
tested with Mint 19 and Ubuntu MATE 18.04 (reproducible in the live environment of both)

lukefromdc:
I cannot duplicate this in the treeview(this is the treeview) and have never had this happen. I use the treeview all the time for sorting by date, and have not had this happen from Caja 1.21 development branch all the way back to my first GTK3 builds of Caja 1.10. I am on Debian Unstable, was on Ubuntu in the 1.10 days.

gm10:
> I cannot duplicate this in the treeview(this is the treeview) and have never had this happen. I use the treeview all the time for sorting by date, and have not had this happen from Caja 1.21 development branch all the way back to my first GTK3 builds of Caja 1.10. I am on Debian Unstable, was on Ubuntu in the 1.10 days.

The only Debian I have at hand right now is MX, which is based on stable, same thing there:
![](https://i.imgur.com/ty9llOw.png)

Since you don't have it, maybe an encoding/code page issue? Didn't test anything but `en_US.UTF-8`. But since it's displaying correctly in one view I'd say it's something in your code. Whatever that view does differently.

Anyway, seems your colleague confirmed this, so good luck with the bug hunt.

sc0w:
I have tested in some **.deb** files, and I confirm the issue in all of them.

lukefromdc:
This is what I get:

![caja_w_debs_9-8-2018](https://user-images.githubusercontent.com/14036792/45256661-9eade500-b367-11e8-9d98-e122a81290ac.png)

gm10:
@lukefromdc this report is about engrampa, not caja

lukefromdc:
[...]

PR #251, opened 2019-02-19
warning: Use of memory after it is freed

rbuj:


lukefromdc:
I agree with pushing this to 1.22, we definately don't want to voluntarily leave a use-after-free in a release version.

raveit65:
np, it's a bugfix and i didn't start to create tarballs.

sc0w:
This PR fixes https://github.com/mate-desktop/engrampa/issues/218

PR #252, opened 2019-02-22
Support .udeb package format

### LINKED ISSUE #246: Support Debian udeb packages
#### Expected behaviour
Using engrampa to open file with ending .udeb should display contents, as with .deb file.

#### Actual behaviour
Fails to open archive, with message: "Archive type not supported".
If archive is renamed with .deb ending, then contents are displayed normally. (udeb and deb are essentially the same file type.)

As above.

#### MATE general version
N/A 

#### Package version
1.20.2

#### Linux Distribution
Debian Stretch 9.7

(none)

Since the two file types (.deb and .udeb) are essentially the same, maybe something similar to the recent fix for handling .a files as .ar would be sufficient for this? https://github.com/mate-desktop/engrampa/commit/499e600d758578eeb8187f92d0c12b36e50f7595

vkareh:
This enables support for udeb (micro-deb) files, which are a package format used during bootstrapping a Debian installation. They have the same mimetype as regular deb files.

Fixes #246


rbuj:
@vkareh 
I suggest you to use "application/vnd.debian.binary-package" instead of "application/x-deb", it is also applicable to .deb file extension.
https://www.iana.org/assignments/media-types/application/vnd.debian.binary-package

vkareh:
@rbuj - good point. I've changed it and it works well, thanks!

I also added another commit that cleans up the list of mime-types and their descriptions

lukefromdc:
I have no idea at all how to test this, as I normally build debs either with Checkinstall or with Debian's files (for their unmodified packages) and dpkg-buildpackage.

vkareh:
@lukefromdc - here's what I used: https://packages.debian.org/stretch/amd64/alsa-utils-udeb/download
 download a random .udeb file from the debian website and use that.

PR #253, opened 2019-02-22
Show version on command line

### LINKED ISSUE #248: [commandline feature request] Please consider adding --version flag to engrampa
Hello,

Doing this:

    engrampa --version

Leads to that:

** (engrampa:11558): CRITICAL **: 19:38:36.664: Failed to parse arguments: Unknown option --version

Would it be possible to add --version to engrampa? --help works already and I can see some options documented.

Use case: I use a little script that, on the commandline, shows which version of which mate-desktop component is installed and available on the local computer system. There are workarounds that I am already using, such as checking for the .so name version string, but consistent "--version" to binaries would be a lot more convenient, so perhaps this could be added. Thanks for reading - please feel free to close this issue at any moment in time.

vkareh:
This adds a `--version` flag to the engrampa cli.

Fixes #248

### DIFF
--- a/data/engrampa.1
+++ b/data/engrampa.1
@@ -63,6 +63,9 @@ Print standard command line options.
 .TP
 \fB\-\-help\-all\fR
 Print all command line options.
+.TP
+\fB\-\-version\fR
+Show the application's version.
 .P
 This program also accepts the standard GTK options.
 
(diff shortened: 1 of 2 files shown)


selectiveduplicate:
Just built and tested- it's working fine. 
Btw should we add this to the man page? :confused:

vkareh:
@selectiveduplicate - good point, I've just added it.

PR #264, opened 2019-03-22
file-utils: avoid out of bound memory access

sc0w:
Fixes Clang static analyzer warning:

```
file-utils.c:453:6: warning: Out of bound memory access (access exceeds upper limit of memory block)
        if (file_name [last_char] == G_DIR_SEPARATOR)
            ^~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/file-utils.c
+++ b/src/file-utils.c
@@ -445,7 +445,7 @@ const gchar* file_name_from_path(const gchar *file_name)
 	if (file_name == NULL)
 		return NULL;
 
-	if (file_name[0] == '\0')
+	if ((file_name[0] == '\0') || (strlen (file_name) == 0))
 		return "";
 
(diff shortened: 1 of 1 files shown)


sc0w:
> Wait, but this change is for line 448, not 453...

Yes, but the warning can be fixed before, see the clang report:



> As for line 448, doesn't (strlen (file_name) == 0) mean the same as (file_name[0] == '\0')?

oops, yes, but it fixes the warning, so, I think the warning is false positive.

monsta:
Clang says weird things here. If `strlen` is more than 0, then `last_char` is the position of the last non-zero char in the string, that is, right before the terminating 0. Therefore it's ok to access it via `file_name[last_char]`.

PR #266, opened 2019-04-02
Use 'strnlen' instead of 'strlen'

sc0w:
Fixes `flawfinder` warnings:

`(buffer) strlen: Does not handle strings that are not \0-terminated; if given one it may perform an over-read (it could cause a crash if unprotected) (CWE-126).`

https://dwheeler.com/flawfinder

http://cwe.mitre.org/data/definitions/126.html

### DIFF
--- a/copy-n-paste/eggsmclient-xsmp.c
+++ b/copy-n-paste/eggsmclient-xsmp.c
@@ -1144,7 +1144,7 @@ array_prop (const char *name, ...)
   va_start (ap, name);
   while ((value = va_arg (ap, char *)))
     {
-      pv.length = strlen (value);
+      pv.length = strnlen (value, BUFSIZ);
       pv.value = value;
       g_array_append_val (vals, pv);
     }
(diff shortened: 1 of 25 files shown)


sc0w:
in my tests with this PR, the value of `BUFSIZ` is 8192

sc0w:
@vkareh 

> I'm honestly not entirely sure about this... yes, strlen cannot handle non-null-terminated strings, but anything declared as char * will be a null-terminated string by definition.

if you are right, I don't understand the warning of flawfinder, because `strlen` needs `char *`

lukefromdc:
Than I'm guessing we can close this?

sc0w:
I can't demostrate problems with strlen in master (without this PR)

closing...

cwendling:
…also simply using `BUFSIZ` when no actually knowing the size of the allocation does *not* solve the issue if it actually existed, one has to use the *actual* size of the allocation (or less).
Anyway, I think it's a good idea not to merge this.

PR #268, opened 2019-04-06
Help: replace link linkend with xref linkend

rbuj:
Yelp viewer can't open cross-references to other parts of
the manual using link linkend.


selectiveduplicate:
So this started happening after upgrading to DocBook 5!

PR #271, opened 2019-05-10
Update ui.h

tamplan:
Remove hard-coded 'Back' label button


lukefromdc:
Master:
Engrampa_Master

With this PR:
Engrampa_PR

tamplan:
Hi,
I purpose to remove the text of the label because it's a hard-coded string of characters.
You are right than Caja has a text label in the back button but it can be translated.

Use of _("Back") can resolve the problem.

Regards

PR #274, opened 2019-06-09
Update FSF address

tamplan:
Update [FSF address](https://www.fsf.org/about/contact/)


tamplan:
Totally lost in the use of git...
I delete my repository and reclone this.

PR #279, opened 2019-07-08
Give unzip priority over p7zip

### LINKED ISSUE #5: Prefer to open zip files with unzip instead of 7z
7z and also versions of unzip older than 6.10b, don't support utf8 properly. Zip archives created on windows that have files with non-latin filenames show up as garbage.

Here's a sample zip archive that displays this problem:
http://pnboy.pinguix.com/gapan/23-10-2012-b-fasi-eaep.zip

Same problem is with cbz files, which are actually zip files.

Removing/commenting out lines 527 and 530 from src/fr-command-7z.c fixes this and opens zip files with unzip, but it removes the mimetypes from 7z completely.

Ideally, priority should be given to unzip first and if unzip is not present, fall back to 7z, if 7z is installed.

### LINKED ISSUE #102: ZIP archieves containing files with non-ascii names are displayed wrongly
There has been already released patch for file-roller, which replaces p7zip with unzip. Look here https://bugs.launchpad.net/ubuntu/+source/p7zip/+bug/1382106

### LINKED ISSUE #201: Cannot decrypt zip with "№" symbol in password
#### Expected behaviour
Correct operation with any zip-passwords

#### Actual behaviour
`ERROR: Wrong password` when pass contains `№` or `℅` or `€`

Create encrypted  zip-archive with password contains symbol `№`
Try to extract with engrampa, type password, getting error (wrong pass).

cli-utility unzip does this operations well, without errors
#### MATE general version
1.20.1

#### Package version
1.20.0

#### Linux Distribution
Debain Sid, also tested in ubuntu 16.04 and Linux Mint 18.3

### LINKED ISSUE #204: 中文乱码 (Chinese garbled)
解压含有中文的压缩文件时，中文名全是乱码。。

> When extracting compressed files containing Chinese characters, the Chinese names are all garbled.

### LINKED ISSUE #210: Cyrillic support
#### Expected behaviour
correct file names into zip-archives maked in windows

#### Actual behaviour
screenshot at 2018-07-18 09-53-48

any file names with a cyrillic symbols

#### MATE general version
any
#### Package version
any

#### Linux Distribution
any

Please, fix this veerryyy oold trouble with zip ( it is really pain :sob: )
(unar do it correctly, but I don't know how to use unar with engrampa)

Thanks.

### LINKED ISSUE #278: Filename encoding in zip archives is processed incorrectly
#### Expected behaviour
Listing files in .zip as unzip does:
Снимок экрана в 2019-07-01 09-55-54

#### Actual behaviour
Снимок экрана в 2019-07-01 09-56-58

1. Open attached .zip file in engrampa
[Desktop.zip](https://github.com/mate-desktop/engrampa/files/3344256/Desktop.zip)

#### MATE general version
1.20.1

#### Package version
mate-desktop-environment (2018.11.28 linuxmint:19.1/tessa [all])

#### Linux Distribution
Linux Mint 19.1 amd64

n/a

Corresponding p7zip issue looks abandoned.
https://sourceforge.net/p/p7zip/bugs/187/

Can we just use unzip instead?

vkareh:
At the moment p7zip does not process OEM encoding in file names (https://sourceforge.net/p/p7zip/bugs/187/) so instead we prioritize unzip.

Fixes #201 
Fixes #5

### DIFF
--- a/src/fr-init.c
+++ b/src/fr-init.c
@@ -362,6 +362,7 @@ register_commands (void)
 	register_command (FR_TYPE_COMMAND_TAR);
 	register_command (FR_TYPE_COMMAND_CFILE);
 	register_command (FR_TYPE_COMMAND_RAR);
+	register_command (FR_TYPE_COMMAND_ZIP);
 	register_command (FR_TYPE_COMMAND_7Z);
 	register_command (FR_TYPE_COMMAND_DPKG);
 
@@ -375,7 +376,6 @@ register_commands (void)
 	register_command (FR_TYPE_COMMAND_LHA);
 	register_command (FR_TYPE_COMMAND_RPM);
 	register_command (FR_TYPE_COMMAND_UNSTUFF);
-	register_command (FR_TYPE_COMMAND_ZIP);
 	register_command (FR_TYPE_COMMAND_LRZIP);
 	register_command (FR_TYPE_COMMAND_ZOO);
 #if HAVE_JSON_GLIB


sc0w:
it doesn't work here.

anyway, `unzip` only supports "read only", so, I don't like it, it must work maybe with a gsettings key.

and, we have the ability to use `unar` to open `zip` files since https://github.com/mate-desktop/engrampa/commit/c587ae127cf0d6fee6b7e8c21bbe1c903faac13c .

I noticed using `unar` fixes https://github.com/mate-desktop/engrampa/issues/278 , and probably more issues.

vkareh:
@sc0w that's a good point, but I think this should work automatically. For a user to have to change gsettings to open a file is not good UX, maybe we can prioritize unar instead of unzip? Why do we need the gsettings instead of just making it automatic?

sc0w:
@vkareh because `unar` open `zip` files with "read only" mode.

I noticed this PR fixes https://github.com/mate-desktop/engrampa/issues/201 but it doesn't fix the name of files in the other issues you mentioned in https://github.com/mate-desktop/engrampa/pull/279#issue-295531519

and zip files are opened locally in "read / write" mode, this is a good point.

vkareh:
> unzip only supports "read only"

and

> unar open zip files with "read only" mode

and

> zip files are opened locally in "read / write" mode

Those three statements confuse me a bit, which ones open in read-only and which ones open in read/write mode?

This fixes the file name issues for me. I wonder what's different between our configurations. I'll play around with settings to see if I find anything.

sc0w:
> Those three statements confuse me a bit, which ones open in read-only and which ones open in read/write mode?

All are opened in read/write mode, I was wrong, I edited that before. I don't know how it works, but it works.

now I agree with this PR because fixes the issue of password, but I don't understand why the names of files works for you, and it doesn't work here.

awaiting the tests of others...

sc0w:
ok, seeing the code I noticed it uses `unzip` to show the files, and the command `zip` to compress

vkareh:
Thanks for the clarification :)

I still haven't found why it may not work for you for file names... I'll keep trying different things on my side to reproduce the issue

vkareh:
@sc0w, I've edited the list of issues this will fix. All other issues are already closed, and the only other one (#102) seems to display the wrong character set with either p7z, unar, or unzip. So I removed it from this list. We're only left with #5 (feature request to prioritize unzip) and #201 (password encoding).

sc0w:
I noticed one problem with this PR with `.zip` files compressed with password by `7z`

how to reproduce:

without this PR, select some files in caja, right click, compress in `zip`, other options, set password, and create. (it is created by 7z).

install this PR, and try to open, it doesn't work, seems it isn't supported by unzip.

file sample, the pass is 123:

[pass123.zip](https://github.com/mate-desktop/engrampa/files/3381316/pass123.zip)

`   skipping: engrampa_1.20.2-1_amd64.buildinfo  need PK compat. v5.1 (can do v4.6)`

raveit65:
What happens with that PR?
Is someone working on it or can we close it?

raveit65:
@vkareh @sc0w 
Any news?

vkareh:
There hasn't been consensus in a year. Closing for now.

PR #282, opened 2019-07-29
Add zcompress support

### LINKED ISSUE #186: [Wishlist] Add Zstd integration
Much more relevant than ZPAQ would be a [Zstandard](https://en.wikipedia.org/wiki/Zstandard) integration.

Zstd uses Finite State Entropy and scales very good. It is faster faster & better compressing than zlib and at the same time much faster than xz (lzma) while delivering only a slightly worse compression ratio.

benchmark

For further benchmarks on your own hardware you may try [TurboBench](https://github.com/powturbo/TurboBench)

Sources are at https://github.com/facebook/zstd

vkareh:
Continues the work from @stokito to fix merge conflicts...

Fixes #186 
Closes #231


ManuLinares:
Question, I'm using Thunar on Manjaro XFCE, Engrampa Archive Manager 1.24.1

**zst** option appears when selecting single file, but it doesn't when selecting a folder or multiple files?

Issue #283, opened 2019-08-24
engrampa fails to build with new glib2 2.61.2


david-geiger:
hi,

I tried on Mageia Cauldron (linux) to build engrampa 1.22.1 but it fails to build with recent glib2 2.61.2:

> In file included from /usr/include/glib-2.0/glib/galloca.h:32,
                 from /usr/include/glib-2.0/glib.h:30,
                 from actions.c:24:
/usr/include/glib-2.0/glib/gtypes.h:551:8: note: declared here
 struct _GTimeVal
        ^~~~~~~~~
In file included from actions.c:43:
glib-utils.h:69:21: error: conflicting types for 'g_ptr_array_copy'
 GPtrArray *         g_ptr_array_copy             (GPtrArray   *array);
                     ^~~~~~~~~~~~~~~~
In file included from /usr/include/glib-2.0/glib.h:31,
                 from actions.c:24:
/usr/include/glib-2.0/glib/garray.h:141:12: note: previous declaration of 'g_ptr_array_copy' was here
 GPtrArray *g_ptr_array_copy               (GPtrArray        *array,

[build.0.20190812145901.log](https://github.com/mate-desktop/engrampa/files/3536951/build.0.20190812145901.log)

Regards,
David

david-geiger:
The same issue in lxqt-archiver project: https://github.com/lxqt/lxqt-archiver/issues/113

david-geiger:
Thank you! now fixed.

lukefromdc:
I can also verify both the issue and that it is now fixed

PR #284, opened 2019-08-24
glib-utils: Remove unused function - g_ptr_array_copy

rbuj:
closes https://github.com/mate-desktop/engrampa/issues/283

### DIFF
--- a/src/glib-utils.c
+++ b/src/glib-utils.c
@@ -568,22 +568,6 @@ get_time_string (time_t time)
 }
 
 
-GPtrArray *
-g_ptr_array_copy (GPtrArray *array)
-{
-	GPtrArray *new_array;
-
-	if (array == NULL)
-		return NULL;
-
-	new_array = g_ptr_array_sized_new (array->len);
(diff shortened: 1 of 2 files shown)


raveit65:
Looks like this is needed to fix ubuntu 19.10 travis CI builds here https://github.com/mate-desktop/engrampa/pull/288

monsta:
> Out of scope, but every function using the g_ namespace is likely to result in a clash like this one, it's not a good idea to have those; they should use a separate prefix.

That's true... we already had similar one at https://github.com/mate-desktop/libmatekbd/pull/49

lukefromdc:
Works here too, I was out of town for a few days and thus backed up

raveit65:
Ok, i will make new release (1.23/1.22).
Building engrampa fails for fedora 31 and rawhide.
https://copr-be.cloud.fedoraproject.org/results/raveit65/MATE-1.23.x/fedora-31-x86_64/01027088-engrampa/build.log.gz

PR #286, opened 2019-08-25
fr-command-lrzip: Fix empty string check

rbuj:
str is empty: if (*str == '\0') {...}

### DIFF
--- a/src/fr-command-lrzip.c
+++ b/src/fr-command-lrzip.c
@@ -76,7 +76,7 @@ list__process_line (char     *line,
 	fdata->dir = FALSE;
 	fdata->link = NULL;
 
-	if (fdata->name == 0)
+	if (*fdata->name == '\0')
 		file_data_free (fdata);
 	else
 		fr_command_add_file (comm, fdata);


src/fr-command-lrzip.c
```diff
@@ -60,6 +60,10 @@ list__process_line (char     *line,
 	fdata->encrypted = FALSE;
 
 	char *new_fname = g_strdup (file_name_from_path (comm->filename));
+
+	if (!new_fname)
+		return;
```
cwendling: this leaks `fdata`

rbuj: fixed

rbuj:
`gchar * g_strdup (const gchar *str);` duplicates a string. If str is NULL it returns NULL.
https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strdup

cwendling:
> `gchar * g_strdup (const gchar *str);` duplicates a string. If str is NULL it returns NULL.
> https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strdup

As I said, it can only return `NULL` if it was given `NULL`.  And in this case, that can only be the case if `comm->filename` is `NULL` already.  And BTW, as `comm->filename` is passed to `stat()` a few lines up (ans `stat()` doesn't allow `NULL`), it ought to have not been not `NULL` or there's another bug above.

Anyway, my point is that the logic of this code is bizarre, and it should better check all aspects of `new_fname` before performing more computation (`NULL`ness if it matters, and emptyness, again if it matters).

BTW, is this fixing an actual issue, or is this speculation?  I would think the string length check was wrong indeed, but the `NULL` check seems like pure speculation, and actually partly incorrect as the fix is not enough if there really is an issue there.

src/fr-command-lrzip.c
```diff
@@ -76,7 +81,7 @@ list__process_line (char     *line,
 	fdata->dir = FALSE;
 	fdata->link = NULL;
 
-	if (fdata->name == 0)
+	if (strlen (fdata->name) == 0)
```
cwendling: what about the faster:
```suggestion
	if (*fdata->name == '\0')
```

rbuj: done

src/fr-command-lrzip.c
rbuj: `new_fname` can't be `NULL`.

raveit65:
Just rebased to get https://github.com/mate-desktop/engrampa/commit/66076cf781963cd1b781ba8e42d32efcf8fa9180 for fixing ubuntu build.

PR #287, opened 2019-08-25
comparison of integers of different signs: 'int' and 'guint'

rbuj:
Fix compiler warning:
comparison of integers of different signs: 'int' and 'guint' (aka 'unsigned int') [-Wsign-compare]

### DIFF
--- a/src/dlg-prop.c
+++ b/src/dlg-prop.c
@@ -157,7 +157,7 @@ dlg_prop (FrWindow *window)
 
 	uncompressed_size = 0;
 	if (fr_window_archive_is_present (window)) {
-		int i;
+		guint i;
 
 		for (i = 0; i < window->archive->command->files->len; i++) {
 			FileData *fd = g_ptr_array_index (window->archive->command->files, i);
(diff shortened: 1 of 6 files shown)


raveit65:
In which log do you see the build warnings?
cppcheck or clang?

rbuj:
I saw those warnings in a local build after modifying configure.ac:
```patch
$ git diff
diff --git a/configure.ac b/configure.ac
index 3f25dc3..2d4cb13 100644
--- a/configure.ac
+++ b/configure.ac
@@ -48,7 +48,7 @@ AC_SUBST([GTK_LIBS])
 
 dnl ===========================================================================
 
-WARN_CFLAGS="-Wall -Wcast-align -Wtype-limits -Wclobbered -Wempty-body -Wignored-qualifiers"
+WARN_CFLAGS="-Wsign-compare"
 
 for option in $WARN_CFLAGS; do
        SAVE_CFLAGS="$CFLAGS"
```
```shell
$ make clean && ./autogen.sh --prefix=/usr && make &> result.txt
$ grep sign-compare result.txt
```

PR #288, opened 2019-08-27
comparison of integer expressions of different signedness: ‘size_t’

rbuj:


raveit65:
Just rebase to fix hopefully ubuntu build at travis.

PR #293, opened 2019-08-29
comparison of unsigned expression >= 0 is always true

rbuj:
java-utils.c:310:23: warning: comparison of unsigned expression >= 0 is always true [-Wtype-limits]


sc0w:
I see the warning:

` [src/java-utils.c:]: (style) The scope of the variable 'j' can be reduced.`

I think you can put `"int j;"` inside the condition `"if (package != NULL)"`

rbuj:
@sc0w 

Done, thank you.

PR #307, opened 2019-09-11
dlg-prop: set property label in ui file

rbuj:


sc0w:
This PR changes the look of the dialog:

Before:



Later:



Please keep the same look as before.

raveit65:
Using `<property name="halign">start</property>` should bring back previour Look&Feel.

rbuj:
See alignment in _Property Windows_ on _GNOME Human Interface Guidelines_:
https://developer.gnome.org/hig-book/2.32/hig-book.html#property-windows

sc0w:
This is MATE, not GNOME :)

Traditional desktop with traditional look.

`caja`, `atril`, `eom` properties dialog have the same classical look.

If you like to add this new look, I think the commit message needs to add description of it (I don't like it)

rbuj:
I can be wrong, as far I know MATE (MATE2) uses  GNOME 2 hig ([Human Interface Guidelines](https://en.wikipedia.org/wiki/Human_interface_guidelines)):

GNOME Human Interface Guidelines 2.0 by the GNOME Documentation Project
Copyright © 2005 Sun Microsystems

https://gitlab.gnome.org/GNOME/gnome-devel-docs/commit/bfa72144e04ae163c11dd8e6047638d0141c4ac0



It's the same screenshot included in [GNOME 1.0 HIG](http://interface.free.fr/Archives/hig-gnome-1.0.pdf) (Page 16)

GNOME Documentation Style Guide V1.0 by The GNOME Usability Project
Copyright © 2002 Sun Microsystems

raveit65:
I don`t care about what some has written somewhere, especially if it comes from gnome.
The result of a wrong design decision from gnome to invent the wheel new and and ship gnome-shell, was the borne of MATE desktop.
Without several design decisions from gnome you would never do PRs here, right?

So, the questions is more does changing the default Look&Feel improve something?
Do we have existing bugs in the last 8 years which can be fixed with a new design?
Does a change introduce another changes in other packages?
Personal i don't really care about.
But if a member of core team (who did maintain engrampa in the last years) disagree with a new design, why not follow him?

Last question, do you really think the existing Look&Feel is bad?

rbuj:
I think that the person who wrote the hig did it for some reason, if the name of the property and the value are closer, it facilitates the reading of the line. I do not remember if I read it in this guide or in another.

I didn't think it was such a major change, for that reason I didn't include it in the commit title. If you think this PR adds a new feature, it would have to be discussed in another thread, so as not to hinder the progress of code improvement.

sc0w:
> If you think this PR adds a new feature, it would have to be discussed in another thread, so as not to hinder the progress of code improvement.

In all the PRs if the authors don't mentioned it, I always expect the same look and behavior as before.

If not, it must be mentioned in PR into the commit mesagge, or into  new commit.

Personally I don't like this new look, my eyes can't see quikly all the properties, for me the human guidelines in gnome, are inhuman guideliness.

I think you can make new PR with that new look, and we can discuss there if other team members are agree.

btw, I appreciate all your contributions to engrampa, you are doing a great job :)

rbuj:
New PR https://github.com/mate-desktop/engrampa/pull/326

PR #308, opened 2019-09-11
dlg-prop: dialog should not be resizable

rbuj:


sc0w:
personally, I don't like to loss features. If the name file, or the location is very long, sometimes maybe it is useful to the user to resize the dialog for some reasons.

lukefromdc:
As a rule I don't tend to favor making things non-resizable due to potential
issues with small screens (e.g old 600x1024 netbooks). I would in fact prefer
that it always be possible to resize things down for that reason. A workaround if
you run compiz is to go to expo mode and move the window until the part you 
need is onscreen after returning to that workspace. Not blocking this, just stating
my own preference here

vkareh:
Good point, I was thinking resizing up to fit the file name (but it already does that), and didn't consider resizing down to fit the screen... Based on that, I think we should close this, since the current behavior is not broken. Any other thoughts/opinions?

sc0w:
If I am right, one member agree versus three members don't agree with this PR.

If there is no more votes, I think we can close the PR.

raveit65:
I like to have more options in general, so why removing a feature?

PR #309, opened 2019-09-11
dlg-prop: show the exact number of bytes as part of the size

rbuj:


raveit65:
@sc0w 
You like that?

sc0w:
It makes sense, because caja shows too the value in bytes into the properties dialog.

btw, I don't understand why I can't see the "**,**" separator like the @lukefromdc screenshot

raveit65:
@sc0w 
Did you test it with another theme?
With Menta:
Bildschirmfoto zu 2019-09-17 08-40-23

sc0w:
no, it doesn't work with menta (the thousands separator)

raveit65:
@sc0w 
Do you think it is a issue with PR or can we merge it?

sc0w:
not related, merging...

PR #322, opened 2019-09-18
fr-command-tar: fetch dates with strptime

rbuj:


rbuj:
@vkareh I think that it's fixed now

raveit65:
@vkareh I think it is expected that properties dialog looks like before. 
So, it is (only) a code simplification.

PR #326, opened 2019-09-21
archive properties dialog: make properties reading easier

rbuj:
Removing the additional separation between the name of the
property and its value, it makes properties reading easier.

### Before

Captura de Pantalla 2019-09-21 a les 23 56 03

### After

Captura de Pantalla 2019-09-21 a les 23 35 26

### DIFF
--- a/src/ui/properties.ui
+++ b/src/ui/properties.ui
@@ -73,7 +73,7 @@
               <object class="GtkLabel" id="p_name_label_label">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
-                <property name="halign">start</property>
+                <property name="halign">end</property>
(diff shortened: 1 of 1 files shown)


rbuj:
I just resolved merge conflicts

sc0w:
Personally I prefer the look of master, for me the new look is weird and I can't see all the properties quickly

lukefromdc:
I also prefer the old look

sc0w:
by the moment, member votes: two versus two

@mate-desktop/core-team please vote

raveit65:
I don't get why i should have different alignment in a excel sheet with 2 columns.
So, think then gnome design rules about using a grid are wrong!
+1 for old look&feel :)

raveit65:
Most team members like the old design. Can we close PR now?

PR #333, opened 2019-10-01
Check if a mime-type is an alias in compute_supported_archive_types

rbuj:
For instance, application/x-lha is an alias of application/x-lzh-compressed,
mime_type_desc.mime_type should only store non duplicated primary mime-types,
to avoid duplicates in the warehouse.


raveit65:
Runs since 10 days without any problem.

Issue #347, opened 2019-12-30
Can't open .ARC files and .PAK files with CAJA/engrampa


InfoLibre:
#### Expected behaviour

Extracting files from .ARC and .PAK archives.

#### Actual behaviour

I can't extract files.

#### Steps to reproduce the behaviour

In CAJA, right-click on the .ARC or .PAK files. The menu **Extract here** is missing.

[MM34.PAK.zip](https://github.com/mate-desktop/engrampa/files/4010932/MM34.PAK.zip)

[EGACHESS.ARC.zip](https://github.com/mate-desktop/engrampa/files/4010937/EGACHESS.ARC.zip)

#### MATE general version

1.22.2

#### Package version

1.22.2

#### Linux Distribution

Linux Mint MATE 19.3

Petrouchka:
I can extract files from .ARC archive with Midnight Commander.

Petrouchka:
I can extract files from .PAK archive with unrar package. Can unrar be used by CAJA ?

raveit65:
Can you please file out a report against your distro at linuxmint bugreporting system and ask there for help, or use Linuxmint forum.
Github is a developing and bugeporting tool, but not user support help line.
Thank you.

Petrouchka:
@raveit65 If CAJA can't open .PAK and .ARC files, it's not a problem with how I use it but it's a bug. Dependencies between CAJA/engrampa and other softwares that can open .PAK and .ARC are missing. If you don't know how to solve this bug in CAJA, why are you closing it without any explanation ?
I take time to report it and your answer is contemptuous.

raveit65:
> #### Link to downstream report of your Distribution

It is very contemptuous and respect-less that you don't honour our template and open a bugreport against your distro.
Anyway, re-opened and ignored by me.

Petrouchka:
A opened a bugreport in my distro : https://github.com/linuxmint/linuxmint/issues/188 but they closed it too. Anyway, thank you rbuj for your help.
With my old ARC files, unar had a bug too. I wrote to macpax.com and they solved it and released a new version https://github.com/MacPaw/XADMaster/releases/tag/v1.10.7

zhuyaliang:
So can we close this issue

PR #348, opened 2019-12-31
Add ARC archive support

rbuj:
You can use EGACHESS.ARC file for testing from #347


raveit65:
@rbuj 
Can you please update PR to fix merge conflict?

rbuj:
@raveit65 done

rbuj:
@lukefromdc install `nomarch` package and run the commands below:
```shell
$ sudo dnf install -y nomarch
$ sudo update-mime-database /usr/share/mime
```

lukefromdc:
Can't use the first of those exact commands: we don't have DNF in Debian. I will have to look for nomarch in Synaptic and see if we have it or a similar package.

raveit65:
> Can't use the first of those exact commands: we don't have DNF in Debian. I will have to look for nomarch in Synaptic and see if we have it or a similar package.

Dnf is a command line tool in fedora, similar to apt or aptitude in debian.
A lot of options are the same, eg.
dnf search ...... = aptitude search ......
If you know to use apt or aptitude , you should able to use dnf, or the other way around.

PR #353, opened 2020-01-18
autoconf: get zstd mime type using libmagic if enabled

rbuj:
Test:
```shell
$ base64 /dev/urandom | head -c 1048576 > test.txt
$ zstd test.txt
$ engrampa test.txt.zst
```


monsta:
Since file/libmagic versions before 5.38 reported the wrong type, maybe just rely on 5.38 and use `application/zstd` always?

rbuj:
Some distros don't ship file 5.38 like Fedora 30 (file 5.36) or Fedora 31 (file 5.37).

```shell
[robert@localhost libmatekbd]$ cat /etc/redhat-release 
Fedora release 30 (Thirty)
[robert@localhost libmatekbd]$ file -v
file-5.36
magic file from /etc/magic:/usr/share/misc/magic
```
[RPM Find Query](https://rpmfind.net/linux/rpm2html/search.php?query=file&submit=Search+...&system=fedora&arch=x86_64)

raveit65:
Can you add a note in configure.ac please, that this fix can be removed with file-5.38?

rbuj:
@raveit65 done

monsta:
```
$ grep MAGIC_VERSION /usr/include/magic.h 
#define MAGIC_VERSION		530	/* This implementation */
```
Maybe we can utilize this somehow?

monsta:
Weird, in my Debian Testing there's no zstd type in `/usr/share/mime` at all. Running `update-mime-database` had no effect. So Caja says `application/octet-stream` in the archive properties.

Piraty:
how about handling libmagic's returned mime types at runtime instead?
```
else if (is_mime_type (comm->mime_type, "application/x-zstd") || is_mime_type (comm->mime_type, "application/zstd") ) {
```

Piraty:
That commit broke crosscompilation as well, as you're using the target compiler

PR #360, opened 2020-01-31
fr-window: duplicated code

rbuj:


raveit65:
Which icon is affected in engrampa window?

rbuj:
The icons shown in the file list (mime types).

PR #366, opened 2020-02-15
Update authors

rbuj:
```
$ git diff --word-diff=color --word-diff-regex=. master
```


raveit65:
@rbuj 
I agree with @vkareh to remove the noreply.github.com adresses.
In general all addresses should be valid, IHMO.

rbuj:
The `noreply.github.com` contains the github username as noted in
https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address.

In case the user decides to keep private the actual email address, the username on github helps to identity the user on github preventing identity theft.

raveit65:
But we can identify this user as author.
You can also write XXX or xyz as author name, it's just the same as noreply.github.com

vkareh:
> In case the user decides to keep private the actual email address, the username on github helps to identity the user on github preventing identity theft.

Then can you regexp replace the noreply email to their github page? Otherwise users will see a mailto link that points to a noreply address

rbuj:
@vkareh done

PR #373, opened 2020-03-30
egg: silent 'Unused parameter' warnings

sc0w:
This is another way to silent the warnings, with (void) instead the gnu macro, I think it will be more portable

### DIFF
--- a/copy-n-paste/eggsmclient-xsmp.c
+++ b/copy-n-paste/eggsmclient-xsmp.c
@@ -445,6 +445,8 @@ sm_client_xsmp_end_session (EggSMClient         *client,
 			    EggSMClientEndStyle  style,
 			    gboolean             request_confirmation)
 {
+  (void) style; /* unused parameter */
+
   EggSMClientXSMP *xsmp = (EggSMClientXSMP *)client;
   int save_type;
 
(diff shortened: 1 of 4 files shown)


raveit65:
@sc0w 
The warnings should be silent with using mate-common-1.24.1
https://github.com/mate-desktop/engrampa/commit/57503bb8b85c10c431000a35628885c9130272a1
Do you agree with using
- `MATE_COMPILE_WARNINGS = yes` as default in configure?
- for travis at master branch `MATE_COMPILE_WARNINGS = maximum`?
- all other branches at travis use the default=yes ?

I setup this already for mate-desktop repo.

sc0w:
> The warnings should be silent with using mate-common-1.24.1

yes, confirmed, 502 warnings of unused parameters were removed in master

anyway, probably I will activate that warnings again at travis only for debian, it gives me useful information reviewing PRs for me

> Do you agree with using

yes, no problem

raveit65:
> anyway, probably I will activate that warnings again at travis only for debian, it gives me useful information reviewing PRs for me

Np , didn`t know that it is possible to enable it for one distro, via CFlags ?

sc0w:
> Np , didn`t know that it is possible to enable it for one distro, via CFlags ?

probably, I don't know, I will investigate it

PR #375, opened 2020-03-31
actions: expand the comment on about dialog

rbuj:
### Before

Screenshot at 2020-03-31 17-29-13

### After

Screenshot at 2020-03-31 17-40-22

### DIFF
--- a/src/actions.c
+++ b/src/actions.c
@@ -899,7 +899,7 @@ activate_action_about (GtkAction *action,
                    "version", VERSION,
                    "copyright", _("Copyright \xc2\xa9 2001–2010 Free Software Foundation, Inc.\n"
                                   "Copyright \xc2\xa9 2012–2020 The MATE developers"),
-                   "comments", _("An archive manager for MATE."),
+                   "comments", _("Engrampa is an archive manager for the MATE Desktop Environment."),
                    "authors", authors,
                    "documenters", documenters,
                    "translator-credits", _("translator-credits"),


rbuj:
The comment is pulled from https://mate-desktop.org/

sc0w:
again, I don't like this change, the about dialogs must to be little, we have more info in the help

rbuj:
@sc0w following the general directions of gtk https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#GtkAboutDialog--comments

Screenshot at 2020-03-31 20-18-50

sc0w:
Thanks to make it simpler.

I think it isn't needed to mention it twice in the dialog:

```
Engrampa

1.24.0

Engrampa is ... .. ..
```

I think is better with just:

`An archive manager for the MATE Desktop Environment.`

instead

`Engrampa is an archive manager for the MATE Desktop Environment.`

or just: `'An archive manager'` because it works in more desktop environments

rbuj:
@sc0w where is the verb in `An archive manager for the MATE Desktop Environment`?

raveit65:
I think it is a good idea to use the name engrampa, a user could be confused and think it is file-roller without that marketing, or not?
I don't really understand what is bad to use the name........

sc0w:
@rbuj why you need a verb? The description doesn't need a verb

In the about dialog the name of program only needs to mention one time, in my opinion. The next sentence is a description, it is redundant to mentioned engrampa again.

But I have no problem if others like to merge it :)

[offtopic] the engrampa icon is a reference to file-roller, I think it is good to change it in the future [/offtopic]

PR #376, opened 2020-04-02
gtk-utils: Remove _gtk_builder_get_widget

rbuj:


raveit65:
What a massive commit. Is it possible to split the commit in single working commits ?

rbuj:
The PR can be split into several commits. In one commit we can remove the helper fuction _gtk_builder_get_widget, in another one we can group the signal handlers, and we can remove builder from structs in another one.

rbuj:
It's only a `git rebase master`.

rbuj:
@raveit65 done

src/ui/batch-add-files.ui
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.22.2 -->
 <interface>
   <requires lib="gtk+" version="3.14"/>
```
raveit65, line 4: Can you check for the same gtk+ version with glade like you did in other UI files?

src/ui/batch-password.ui
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.22.2 -->
 <interface>
   <requires lib="gtk+" version="3.14"/>
```
raveit65, line 4: Same here.

rbuj:
@raveit65 done. All ui files require same gtk+ version from now.

raveit65:
@rbuj 
Can you please explain why replacing _gtk_builder_get_widget with a define is needed?
Well, this a code rewrite which does the same thing as before
What is the advantage?
I ask because other people starts to do the same in other repos without given any reason, only because your PR.
I mean doing this only because we can do that is not really a reason:)
So, please can you explain this a bit?

rbuj:
`_gtk_builder_get_widget` is a helper function which wraps the call to [gtk_builder_get_object ()](https://developer.gnome.org/gtk3/stable/GtkBuilder.html#gtk-builder-get-object). Here, it only converts the returned GObject pointer to a GtkWidget pointer.

```C
GtkWidget *	
_gtk_builder_get_widget (GtkBuilder *builder,	
			 const char *name)	
{	
	return (GtkWidget *) gtk_builder_get_object (builder, name);	
}
```

In other projects, a similar helper function can reduce the number of arguments of `gtk_builder_get_object` when the pointer to a GtkBuilder has always the same name in the source file.

```C
#define GET_LABEL(x) (GTK_LABEL (gtk_builder_get_object (builder, (x))))
```

Being GtkWidget a base class, sometimes we might be more interested to access directly to a derived type of GtkWidget or don't want to box the GObject pointer, in order to reduce boxing/unboxing we can use:

```C
#define GET_LABEL(x) (GTK_LABEL (gtk_builder_get_object (builder, (x))))

<cut>
gtk_label_set_label (GET_LABEL ("pw_password_label"), text);
<cut>
```

Instead of:

```C
<cut>
GtkWidget  *label;

label = _gtk_builder_get_widget (builder, "pw_password_label");
gtk_label_set_label (GTK_LABEL (label), text);
<cut>
```

rbuj:
The next step could be to delete the object lookup by declaring an object derived from the GtkDialog, and bind the widgets and the callbacks:

Gtk tutorial: Adding a side bar
https://developer.gnome.org/gtk3/stable/ch01s04.html

PR #392, opened 2020-06-12
Do not use stock icons in ui files

rbuj:


rbuj:
I could not find any replacement for these two stock icons, and it's recommended to use only the label:
- [GTK-STOCK-APPLY](https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html#GTK-STOCK-APPLY:CAPS)
- [GTK-STOCK-OK](https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html#GTK-STOCK-OK:CAPS)

raveit65:
We decided in the past not to fix deprecations in menus or buttons if it isn't possible to do that without removing icons. Because this is one of the main visual different to Gnome, and it is a reason to use Mate.
But maybe @sc0w knows a solution?

lukefromdc:
We could of course copy them into mate-icon-theme under another name.

On 6/23/2020 at 9:26 AM, "Robert Antoni Buj Gelonch"  wrote:
	I could not find any replacement for these two stock icons, and it's
recommended to use only the label: 

 	*GTK-STOCK-APPLY 	*GTK-STOCK-OK

rbuj:
These icons were removed two years ago from the development branch (mainline):
https://github.com/GNOME/gtk/commit/806c659efe3a8f2ebc4a5201f7152636edf3bb49#diff-6a13c6ff6eb762cc72ee27ce157b4362

But they are still available in branch gtk-3-24:
https://github.com/GNOME/gtk/tree/gtk-3-24/gtk/icons/22x22/actions
- gtk-apply.png
- gtk-cancel.png
- gtk-no.png
- gtk-ok.png
- gtk-yes.png

sc0w:
@rbuj 

> I could not find any replacement for these two stock icons, and it's recommended to use only the label:

We don't accept changes losing icons, keep it with no changes if you don't know how to replace them

rbuj:
The look up in the stock items database is no longer used for ui files and it has been broken down into two components, the label and the icon.

raveit65:
I don't think that we use gtk-4.0 branch in the next years, so it is Ok to use icon names from latest gtk-3.0 branch.

raveit65:
@sc0w 
Issue with missing icons is fixed. 
Do you have any concerns left or can i merge it?

raveit65:
Squashed and Done.

PR #396, opened 2020-06-21
Allow preferring unzip over 7z, as it supports OEM encodings (#5)

alkisg:
Hi, this PR addresses #5, i.e. it allows preferring unzip over 7z, because unzip supports OEM encodings and 7z doesn't and won't.

For example, if a Greek (or any non-English) user opens Windows explorer, right clicks on a folder and selects "Create .zip file", that .zip file can't be properly uncompressed with 7z, as it's using the OEM (DOS) encoding.

To tell unzip which OEM encoding was used, the user is expected to define the following environment variables:

```shell
export UNZIP="-O cp737"
export ZIPINFO="-O cp737"
```
Thus this patch checks if UNZIP is defined, and if so, it registers unzip before 7z.

### DIFF
--- a/src/fr-init.c
+++ b/src/fr-init.c
@@ -363,9 +363,14 @@ register_commands (void)
 	 * have higher priority over commands that can only read the same
 	 * format, regardless of the registration order. */
 
+	/* Allow preferring unzip over 7z, as it supports OEM encodings (#5) */
+	gboolean prefer_zip = (getenv("UNZIP") != NULL);
+
 	register_command (FR_TYPE_COMMAND_TAR);
 	register_command (FR_TYPE_COMMAND_CFILE);
 	register_command (FR_TYPE_COMMAND_RAR);
+	if (prefer_zip)
+		register_command (FR_TYPE_COMMAND_ZIP);
 	register_command (FR_TYPE_COMMAND_7Z);
 	register_command (FR_TYPE_COMMAND_DPKG);
 
@@ -380,7 +385,8 @@ register_commands (void)
 	register_command (FR_TYPE_COMMAND_LHA);
 	register_command (FR_TYPE_COMMAND_RPM);
 	register_command (FR_TYPE_COMMAND_UNSTUFF);
-	register_command (FR_TYPE_COMMAND_ZIP);
+	if (!prefer_zip)
+		register_command (FR_TYPE_COMMAND_ZIP);
 	register_command (FR_TYPE_COMMAND_LRZIP);
 	register_command (FR_TYPE_COMMAND_ZOO);
 #if HAVE_JSON_GLIB


alkisg:
Note for end users: one way to define these environment variables is to create a /usr/local/bin/engrampa file, with the following contents:

```shell
#!/bin/sh
# Set the correct OEM encoding for engrampa

export UNZIP="-O cp737"
export ZIPINFO="-O cp737"
exec /usr/bin/engrampa "$@"
```

Replace "cp737" with the appropriate [DOS encoding for your country](https://en.wikipedia.org/wiki/Windows_code_page#DOS_code_pages).

sc0w:
it doesn't work here, `view -> last output` says me `zip` files are opened with `p7zip`

anyway, in my point of view, the user must can decide which prefer to use, gsettings key is needed

and unzip have issues opening files with password, see https://github.com/mate-desktop/engrampa/pull/279#issuecomment-510414238

alkisg:
I'm attaching screenshots that prove that it works.
I used the win10test.zip file that I attached in https://github.com/mate-desktop/engrampa/issues/5#issuecomment-303002294.
The "before-patch" screenshot displays what non-English users see when opening windows-generated zip files. It's really important to give them at least a temporary way to open such files.
The "after-patch" screenshots shows the correct output.

Remember that to activate the patch AND to instruct unzip about the correct charset, you need to define the UNZIP environment variable. If you prefer to use a gsetting for this, then the users will have to toggle two things, both the gsetting and the environment variable, which would be very inconvenient.

So, for "after-patch", the command I used was:

```shell
UNZIP="-O cp737" ZIPINFO="-O cp737" engrampa win10test.zip 
```
before-patch

after-patch

unxed:
Why can not engrampa detect appropriate DOS encoding by system locale settings and set environment variables automatically?

sc0w:
we need to do the things easily to beginner users, if it needs additional steps before running engrampa, it is no good idea, engrampa must do the job itself in my opinion.

we need a way to detect the encoding (cp737) inside engrampa code

and add gsettings key to choose between unzip or p7zip

alkisg:
That would be a patch for unzip, not for engrampa. There have been [various efforts](https://bugs.launchpad.net/ubuntu/+source/unzip/+bug/580961) for that for more than 10 years, but none of them became mainstream enough to reach the distributions. But the "-O charset" environment variables are supported.

We only ask from engrampa to allow us to prefer unzip, at least temporarily; engrampa code shouldn't bother with autodetecting encodings at all, as it's a huge task.

alkisg:
As mentioned in https://github.com/mate-desktop/engrampa/issues/5#issuecomment-648410042, I'm closing this PR as unxed's p7zip-based solution is much better.

PR #399, opened 2020-07-12
Avoid memory leak in java utils

shlyakpavel:
g_strndup allocated the memory for the second time without releasing it
I understand my way of fixing the issue is not perfect so I'd love to hear your suggestions on it

### DIFF
--- a/src/java-utils.c
+++ b/src/java-utils.c
@@ -309,7 +309,9 @@ get_package_name_from_class_file (char *fname)
 		for (j = length; (j >= 0) && (end == 0); j--)
 			if (package[j] == '/')
 				end = j;
-		package = g_strndup (package, end);
+		char *package_padded = g_strndup (package, end);
+		g_free(package);
+		package = package_padded;
 	}
 
 	java_class_file_free (cfile);


rbuj:
> g_strndup allocated the memory for the second time without releasing it
> I understand my way of fixing the issue is not perfect so I'd love to hear your suggestions on it

A better approach would be to set the end position instead of making a realloc:
```patch
diff --git a/src/java-utils.c b/src/java-utils.c
index 7fa0ff3..136defe 100644
--- a/src/java-utils.c
+++ b/src/java-utils.c
@@ -309,9 +309,7 @@ get_package_name_from_class_file (char *fname)
                for (j = length; (j >= 0) && (end == 0); j--)
                        if (package[j] == '/')
                                end = j;
-               char *package_padded = g_strndup (package, end);
-               g_free(package);
-               package = package_padded;
+               package[end+1] = '\0';
        }
 
```

raveit65:
@shlyakpavel 
So this can be improved?
See previous comment.

shlyakpavel:
@raveit65 I'm not sure

cwendling:
@rbuj's method is probably better (although it keeps a larger allocation than needed, but IIUC it's quite small, which makes it possibly better than allocating a new + free old).  I'd even go further and clean up the code like so:
```c
	if (package != NULL) {
		int j;
		for (j = length; j >= 0; j--) {
			if (package[j] == '/') {
				/* truncate string */
				package[j] = '\0';
				break;
			}
		}
	}
```
and remove the `end` variable.

Or even
```c
	if (package != NULL) {
		char *last_slash = strrchr (package, '/');
		if (last_slash)
			*last_slash = '\0';
	}
```
and also remove the `length` variable.  Admittedly it might scan more memory (as we don't pass it the length), but that's probably not a biggie, and the overall byte search is probably faster (`strrchr()` is likely taking advantage of several architecture-specific optimizations as well as having been carefully crafted for speed).  This could be avoided by using `memrchr()`, but that's `_GNU_SOURCE`, not sure what's our take on that.

rbuj:
@cwendling your proposals do not work if the class does not belong to any package.

Test open text.jar file using engrampa, drag and drop HelloWorld.class file on engrampa window. 

```
cat <<EOF > HelloWorld.java
class HelloWorld {
    public static void main(String[] args) {
        System.out.println("Hello, World!");
    }
}
EOF
javac HelloWorld.java

cat <<EOF > HelloWorld2.java
package org.mate.text;
class HelloWorld2 {
    public static void main(String[] args) {
        System.out.println("Hello, World!");
    }
}
EOF
javac HelloWorld2.java

jar cf test.jar HelloWorld.java HelloWorld2.java
```

cwendling:
Ah right, the case where there's no `/` in `package`.  Easy enough to fix:
```c
	if (package != NULL) {
		int end;

		for (end = length; end > 0; end--) {
			if (package[end] == '/')
				break;
		}
		package[end] = '\0';
	}
```
or even
```c
	if (package != NULL) {
		int end;

		for (end = length; end > 0 && package[end] != '/'; end--)
			/* nothing */;
		package[end] = '\0';
	}
```

And the second version:
```c
	if (package != NULL) {
		char *last_slash = strrchr (package, '/');
		if (last_slash)
			*last_slash = '\0';
		else
			*package = '\0';
	}
```

rbuj:
The 1st version: The programming style guides say that a bodyless for loop should be rewritten as a while loop to make it easier to read, in this case I think it's necessary to use the sentence continue.

The 2nd version: Why do you  mix declaration and code here? Do not use the same variable in strrchr call / assigment if last_slash == NULL.

raveit65:
Any progress?

shlyakpavel:
> Any progress?

Looks like there's a club of clever guys inventing thousands of ways to improve one line of the code :D

PR #409, opened 2020-10-05
Add Electronic Publication (EPUB) support

### LINKED ISSUE #408: Cannot unzip .epub

rbuj:
closes #408 

Test: http://files.infogridpacific.com/epub3/don-quijoti-epub3.epub
```
$ file -b --mime-type don-quijoti-epub3.epub
application/epub+zip
```

Run the commands below before opening don-quijoti-epub3.epub from caja:
```shell
$ sudo update-mime-database /usr/share/mime
```


raveit65 (approved):
LGTM,
works like charm and i see engrampa in open-with menu of caja.

raveit65:
Should we consider this as bugfix for 1.24?
Or new feature for next mayor release?

rbuj:
imho it's a new feature.

robert-pearson:
I tried it on my system (Ubuntu MATE 18.04.5 AMD 64-bit). Still fails - not a fix. The difference between a program and an algorithm is that a program works one time in one environment. An algorithm is required to work every time in every environment.
I have tried 19.04, 19.10 and 20.4. the 19.X breaks code which works in 18.X. 20.4 will not even install (the drivers for WiFi network are missing or broken).

raveit65:
Mime-types are updated after installation?
You can ask ubuntu-maintainer to add it to your distro before next Mate major release.

akshaykhoje:
@rbuj I tried your suggested fix here to extract an epub file. However, it doesn't work for me. I tried it after a reboot too, but the same result.

However, since these two commands were new to me so I learned something today. Thanks for that!

But yeah, the issue remains unresolved for me.

I have the following software versions installed :
**OS** : Debian 11 Bullseye
**DE** : MATE 1.24.1
**Engrampa** : 1.24.1
**Caja** : 1.24.0

robert-pearson:
The specification for an ePub file is that it MUST be a zip file. Change
the extension to .zip and then Engrampa can expand it. The coder does not
check to see if the file is a zip file, he relies only on the extension to
the file name. If he bothered to actually open the file, he would see that
it is a zip file.
If the .zip file does not open, perhaps it is corrupt. Try downloading it
again.
Robert Pearson

akshaykhoje:
Yup, it works! Thanks @robert-pearson!

Can someone give me directions to map the epub file extensions to zip. I would like to fix this for convenience.

If I'm framing this wrong, feel free to correct me. I don't know exactly how this is to be put, but I guess it's clear that I wish to open epub files directly without converting/renaming it to .zip.

PR #410, opened 2020-10-06
fr-archive: Fix memory leak

### LINKED ISSUE #400: Duplicated assignment
https://github.com/mate-desktop/engrampa/blob/d75c80047df558a76c7d276adbedf24e337f1650/src/fr-archive.c#L1918-L1920
It is possible that removing one of the repeating assignments is enough. It as also possible, that the author meant to set another field of xfer_data. Fields source_id, action and uri are not set, sooo.. I really hope someone more familiar with the code investigates it. Have a nice day!

rbuj:
closes #400

### DIFF
--- a/src/fr-archive.c
+++ b/src/fr-archive.c
@@ -1917,7 +1917,6 @@ copy_remote_files (FrArchive     *archive,
 	xfer_data->base_uri = g_strdup (base_uri);
 	xfer_data->dest_dir = g_strdup (dest_dir);
 	xfer_data->update = update;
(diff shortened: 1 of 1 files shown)


cwendling (approved):
LGTM, line is duplicated from 2 lines above, leading to leaking the resource previously assigned.

raveit65 (approved):
LGTM.

lukefromdc:
Late review (time again) but I just tested this and it works without issue here. Anyway, one more successful test before a release containing this.

PR #418, opened 2021-01-23
 @oubayashi glib-utils: add judgement whether base_dir is null in _g_path_get_base_name

oubayashi:


rbuj (commented):
`base_dir` cannot be `NULL`, otherwise it prevents a segment violation as `NULL` cannot be passed to strlen.  Note `strlen` never returns any value less than zero.
```C
const char *
_g_path_get_base_name (const char *path,
                       const char *base_dir,
                       gboolean    junk_paths)
{
        size_t      base_dir_len = 0;
        const char *base_path;

        if (junk_paths)
                return _g_path_get_file_name (path);

        if (base_dir != NULL) 
                base_dir_len = strlen (base_dir);

        if (strlen (path) < base_dir_len)
                return NULL;

	base_path = path + base_dir_len;
        if (path[0] != '/')
                base_path -= 1;

        return base_path;
}
```
POSIX refs
- `size_t strlen(const char *s);` [[REF](https://pubs.opengroup.org/onlinepubs/009696899/functions/strlen.html)]
- `size_t` shall be an unsigned integer type. [[REF](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html)]

PR #421, opened 2021-01-26
rar 6.00: fix listing archive content with encrypted file list

### LINKED ISSUE #374: Cannot open RAR archives with encrypted file list
#### Actual behaviour
Encrypted RAR archives cannot be opened via Engrampa (as well as Gnome file-roller)

1. Install Debian Stable with MATE
2. Install the latest RAR for Linux version [5.90](https://www.rarlab.com/rar/rarlinux-5.9.0.tar.gz)

#### MATE general version
1.20.3

#### Linux Distribution
Debian Stable (32 & 64 bit)

Hello,
When I create an archive within Engrampa (or file-roller) and enable [Encrypt the file list](http://0x0.st/iMMd.png) option I can not open it afterwards and get the following [error](http://0x0.st/iMMn.png).

Everything was fine in versions before 5.80 (so I reverted back to using [5.7.1](https://www.rarlab.com/rar/rarlinux-5.7.1.tar.gz)). That's how [it should look](http://0x0.st/iMM7.png) when you open an archive.

I also started Engrampa from the console to see if there are any errors, but there weren't.

When I extract such archives **in console everything is OK**, so I guess it's not a RAR bug.
I also see there was a [similar problem](https://github.com/mate-desktop/engrampa/issues/3) here long ago.

Archives **created in Windows** and then being opened in Linux **are also OK** and it's really strange.

I contacted the developer ([Eugene Roshal](mailto:dev@rarlab.com)) and he said he needed some more information like what parameters are being passed from/to Engrampa or some other GTK library responsible for that in order to track this issue down.

Could you please provide such info? Probably we should open an issue somewhere in GNOME/GTK bug tracker...

Thank you in advance,
Alex

rbuj:
closes #374

### DIFF
--- a/src/fr-command-rar.c
+++ b/src/fr-command-rar.c
@@ -687,18 +687,10 @@ fr_command_rar_handle_error (FrCommand   *comm,
 	for (scan = g_list_last (comm->process->err.raw); scan; scan = scan->prev) {
 		char *line = scan->data;
 
-		if (strstr (line, "password incorrect") != NULL) {
-			error->type = FR_PROC_ERROR_ASK_PASSWORD;
-			break;
-		}
-
-		if (strstr (line, "password is incorrect") != NULL) {
-			error->type = FR_PROC_ERROR_ASK_PASSWORD;
-			break;
-		}
-
-		if (strstr (line, "wrong password") != NULL)
-		{
+		if ((strstr (line, "Incorrect password") != NULL) ||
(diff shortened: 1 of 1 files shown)


raveit65 (approved):
LGTM,
it works fine and fixes related report about encrypted file list.
I used steps-to-reproduce from https://github.com/mate-desktop/engrampa/issues/374#issuecomment-767663628

PR #424, opened 2021-02-05
build: don't add twice the file engrampa.appdata.xml to CLEANFILES

rbuj:
The variable `appdata_DATA` is set to `engrampa.appdata.xml`


raveit65 (approved):
LGTM

PR #437, opened 2021-10-21
Fix memory leaks

rbuj:


src/fr-command-tar.c
```diff
@@ -140,8 +140,13 @@ process_line (char     *line,
 	}
 	g_free (name);
 	name = g_filename_from_utf8 (fdata->original_path, -1, NULL, NULL, NULL);
-	if (name)
-		fdata->original_path = name;
+	if (name) {
+		if (fdata->free_original_path)
+			g_free (fdata->original_path);
```
mbkma, line 145: redundant null check

mbkma:
other than that its fine

src/fr-command-tar.c
rbuj: done

mbkma (approved):
LGTM

src/fr-command-tar.c
rbuj: @mbkma `fdata->original_path` is not NULL and points to `fdata->full_path`.

raveit65:
@rbuj 
Can you please trigger a deployment at travis?

raveit65 (approved):
LGTM

PR #440, opened 2021-10-25
glib-utils: escape_str_common should be static

rbuj:
The function is only used in escape_str_common.c and it's called after its implementation.

### DIFF
--- a/src/glib-utils.c
+++ b/src/glib-utils.c
@@ -118,7 +118,7 @@ count_chars_to_escape (const char *str,
 	return n;
 }
 
-char*
+static char*
 escape_str_common (const char *str,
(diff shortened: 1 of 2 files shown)


mbkma (approved):
LGTM

```
$ grep -r escape_str_common src/
grep: src/engrampa-server: binary file matches
src/glib-utils.c:escape_str_common (const char *str,
src/glib-utils.c:	return escape_str_common (str, meta_chars, '\\', 0);
grep: src/glib-utils.o: binary file matches
grep: src/engrampa: binary file matches
src/glib-utils.h:char*               escape_str_common            (const char *str,
```

raveit65 (approved):
LGTM

PR #442, opened 2021-10-26
test-server: Remove unreachable code after calling g_error

rbuj:


mbkma (approved):
LGTM.
From the docs:
`Error messages are always fatal, resulting in a call to G_BREAKPOINT() to terminate the application. This function will result in a core dump; don't use it for errors you expect. Using this function indicates a bug in your program, i.e. an assertion failure.`

raveit65 (approved):
LGTM

PR #453, opened 2022-03-08
Add .pkg support for FreeBSD

HacKurx:
Freebsd has recently changed the extension of its packages [1], so it is necessary to add the .pkg extension in order to open them with engrampa.

[1] https://cgit.freebsd.org/ports/commit/?id=e497a16a286972bfcab908209b11ee6a13d99dc9

### DIFF
--- a/src/fr-init.c
+++ b/src/fr-init.c
@@ -150,6 +150,7 @@ FrExtensionType file_ext_type[] = {
 	{ ".otp", "application/vnd.oasis.opendocument.presentation-template" },
 	{ ".ots", "application/vnd.oasis.opendocument.spreadsheet-template" },
 	{ ".ott", "application/vnd.oasis.opendocument.text-template" },
+	{ ".pkg", "application/x-xz-compressed-tar" },
 	{ ".rar", "application/x-rar" },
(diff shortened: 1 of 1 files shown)


rbuj:
The [.pkg file extension](https://en.wikipedia.org/wiki/.pkg) is used for other file types. In addition, .pkg file extension can have two compression methods on FreeBSD from now, zstd or xz.

HacKurx:
> The [.pkg file extension](https://en.wikipedia.org/wiki/.pkg) is used for other file types.

Mate must privilege the environments which are linked to it and thus where it can be easily installed, that is to say FreeBSD for this extension.

> In addition, .pkg file extension can have two compression methods on FreeBSD from now, zstd or xz.

zstd is available but not used for packages, if you look at [the code](https://github.com/freebsd/freebsd-ports/blob/4048503392dbd565c59143a0298296a3fdbe943a/Mk/bsd.port.mk#L2229) you will see that its part is commented.
```
#.if ${OSVERSION} > 1400000
#PKG_COMPRESSION_FORMAT?=	tzst
#.else
```

lukefromdc:
We need someone who has OpenBSD installed to have a look at this, been sitting here ages

PR #454, opened 2022-03-14
glib-utils: redundant condition

rbuj:
```
$ cppcheck --enable=all --quiet . &> cppcheck.log
$ grep -A 2 redundantCondition cppcheck.log 
src/glib-utils.c:366:24: style: Redundant condition: If 'EXPR == ' '', the comparison 'EXPR != 0' is always true. [redundantCondition]
 while ((*line == ' ') && (*line != 0))
                       ^
```

### DIFF
--- a/src/glib-utils.c
+++ b/src/glib-utils.c
@@ -363,7 +363,7 @@ eat_spaces (const char *line)
 {
 	if (line == NULL)
 		return NULL;
-	while ((*line == ' ') && (*line != 0))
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Builds and runs fine, oh yeah this is redundant as it is not possible to satisfiy the first condition without automatically satisfying the second. If a variable equals one value it also cannot equal another different value so for this commit to create a problem should be entirely impossible

PR #456, opened 2022-07-24
fr-window: fix -Wincompatible-pointer-types warning

rbuj:
```
fr-window.c:5779:33: warning: incompatible pointer types passing 'GtkWidget *' (aka 'struct _GtkWidget *') to parameter of type 'GtkLabel *' (aka 'struct _GtkLabel *') [-Wincompatible-pointer-types]
        gtk_label_set_mnemonic_widget (window->priv->location_label,
                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/gtk-3.0/gtk/gtklabel.h:129:70: note: passing argument to parameter 'label' here
void     gtk_label_set_mnemonic_widget            (GtkLabel         *label,
                                                                     ^
```

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -5776,7 +5776,7 @@ fr_window_construct (FrWindow *window)
 	gtk_entry_set_icon_from_icon_name (GTK_ENTRY (window->priv->location_entry),
 				       GTK_ENTRY_ICON_PRIMARY,
 				       "folder");
-	gtk_label_set_mnemonic_widget (window->priv->location_label,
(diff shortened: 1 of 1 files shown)


zhuyaliang (approved):
Remove warning

lukefromdc (approved):
Builds and runs fine. This solution has worked many times before for the same warnings, and in some cases those warnings came with a side order of real problems even if this time around compilers seem to figure out how to build the code right. That should not be relied upon so good to fix this right

PR #457, opened 2022-07-24
fr-command-rar: Parameter 'line' can be declared with const

rbuj:
```
cppcheck --enable=all --quiet . &> cppcheck.log
grep constParameter cppcheck.log 
```

### DIFF
--- a/src/fr-command-rar.c
+++ b/src/fr-command-rar.c
@@ -151,7 +151,7 @@ mktime_from_string (const char *date_s,
 }
 
 static gboolean
-attribute_field_with_space (char *line)
+attribute_field_with_space (const char *line)
 {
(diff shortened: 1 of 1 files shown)


zhuyaliang (approved):
Using const decoration is safer

lukefromdc (approved):
fr-command-rar.o builds fine with this installed, but I do not have rar installed due to the shareware/registration license which I do not consent to. I have unrar-free installed but no rar archives to test this for function with.

lukefromdc:
This was written by one team member and tested by another as well as myself. Code wise no reason this should break, so will proceed to merge as a 3ed review (though highly desirable) seems unlikely these days

PR #458, opened 2022-07-24
file-utils: remove unused functions

rbuj:


zhuyaliang (approved):
Agree to remove useless code

lukefromdc (approved):
Builds and runs fine, removing a function that is used should cause an immediate build failure

PR #469, opened 2022-12-15
fix: Two BUG 

snowsi:
1. Encrypted documents in. 7z format cannot delete locked files and folders
    Step: Double click to open a. 7z encrypted document, select a file, and right-click to delete it.

    Because the rm command is executed after the delete command is executed, the error output of the delete command is 
    released, and the_ command_ 7z_ handle_ Error cannot be caught in error

2. Encrypted documents in all formats cannot be successfully pasted to the specified directory after copying and cutting 
    encrypted files and folders
    Step: Double click to open the. 7z encrypted document, select a file, right-click to copy, cut, and then right-click to paste it 
    into the package folder

    Because the error information of the last command was not cleared during false loading, the error information of the last 
    command was processed during error processing


src/fr-command-7z.c
```diff
@@ -393,13 +393,30 @@ fr_command_7z_add (FrCommand     *comm,
 	fr_process_end_command (comm->process);
 }
 
+static void
+process_line__delete (char     *line,
+		      gpointer  data)
+{
+	FrCommand *comm = FR_COMMAND (data);
```
zhuyaliang, line 400: Unused variable 'comm'

src/fr-window.c
```diff
@@ -6317,14 +6317,34 @@ fr_window_archive_add_dropped_items (FrWindow *window,
 				      window->priv->volume_size);
 }
 
+static void
+remove_data_free (GList *rdata)
+{
+	g_list_free_full(rdata, g_free);
```
zhuyaliang, line 6323: Add space

lukefromdc: Another trivial fix, then this conversation can be resolved and this should be mergable

src/fr-command-7z.c
lukefromdc: This should be trivial to fix

src/fr-window.c
zhuyaliang: @lukefromdc If the author still doesn't reply after two days, I will merge this PR

zhuyaliang:
@lukefromdc Any further questions? I'm going to merge

lukefromdc:
No further questions, feel free to proceed

Issue #414, opened 2020-11-19
The "Save As..." for tar related archive file would fail, since tar 1.29


Lemuel-Liu:
#### Expected behaviour
A tar related archive file can be saved as another one, successfully.

#### Actual behaviour
It would fail, giving following tips.

tar: '--keep-newer-files' cannot be used with '--keep-old-files'
Try 'tar --help' or 'tar --usage' for more information.
![image](https://user-images.githubusercontent.com/74695299/99632167-bd47b480-2a77-11eb-9b91-7cc1c2226907.png)

#### Steps to reproduce the behaviour
1. create an archive file with tar;
2. open it with engrampa;
3. select "Save As..." to create another archive file(still tar related);
make sure the version of  tar is above 1.29

#### MATE general version
mate-desktop                           1.24.0-2

#### Package version
engrampa                               1.24.0-2

#### Linux Distribution
ubuntu-mate 20.04

Lemuel-Liu:
Since tar 1.29, it improves option sanity checking. And we can't use "--keep-newer-files" and "--keep-old-files" at the same time.
But when we select "Save As..." in engrampa, it will actually satisfy following two conditions simultaneously. Though "Extract..." would not.
fr_command_tar_extract:
	if (! overwrite)
		fr_process_add_arg (comm->process, "-k");
	if (skip_older)
		fr_process_add_arg (comm->process, "--keep-newer-files");

I think the judgement for "skip_older" could be changed a little. Add an "else" before "if" to make sure their mutual exclusion.
	if (! overwrite)
		fr_process_add_arg (comm->process, "-k");
	else if (skip_older)
		fr_process_add_arg (comm->process, "--keep-newer-files");

Or we could change the values of "overwrite" and "skip_older", passed from "Save As..."?

PR #470, opened 2023-03-15
Fix archive file save as function

zhuyaliang:
Fix archive file save as functio. When saving as, it will check whether the two archive files are the same

#414


lukefromdc:
OK, I see the issue here: "save as" is ALWAYS greyed out

lukefromdc (approved):
This works, note that it can only "save as" when you are making an archive containing at least two files and thus may want to name the new archive something other than the directory name. Assuming that is intended behavior.

lukefromdc:
Memory management looks clean here as well

PR #471, opened 2023-03-16
Debug is enabled by default

zhuyaliang:
We should open debug by default, which is convenient for users to debug and more helpful for us to fix bugs


lukefromdc (commented):
Does this make Engrampa significantly heavier (cpu usage/RAM usage) under any runtime scenario?

zhuyaliang:
@lukefromdc 
```
[test@localhost engrampa]$ grep -rin MATE_ENABLE_DEBUG src/
src/fr-command-cfile.c:79:#ifdef MATE_ENABLE_DEBUG
src/fr-process.c:709:#ifdef MATE_ENABLE_DEBUG
src/fr-process.c:915:#ifdef MATE_ENABLE_DEBUG
src/fr-window.c:2747:#ifdef MATE_ENABLE_DEBUG
src/fr-window.c:3003:#ifdef MATE_ENABLE_DEBUG
src/fr-window.c:3041:#ifdef MATE_ENABLE_DEBUG
src/fr-window.c:7793:#ifdef MATE_ENABLE_DEBUG
src/fr-archive.c:666:#ifdef MATE_ENABLE_DEBUG
src/fr-archive.c:890:#ifdef MATE_ENABLE_DEBUG
src/fr-command.h:63:#if MATE_ENABLE_DEBUG
src/glib-utils.c:426:#ifdef MATE_ENABLE_DEBUG
src/fr-command.c:82:#if MATE_ENABLE_DEBUG
src/java-utils.c:223:#ifdef MATE_ENABLE_DEBUG
```
little debugging, That will not have a significant impac

lukefromdc (commented):
Just tested this, when running engrampa from the terminal displays a lot of debugging info by default. At least they don't show up in .xsession-errors or syslog, but I suspect this may confuse new users or create problems when engrampa is used in terminal.

Someone building locally can easily disable this but most end users cannot. Not sure this is a good idea but not hard-blocking

zhuyaliang:
Yes, some users prefer clean terminals. We should set an option for users to choose whether to turn debugging on or off, such as this

```engramma --debug``` or ```export ENGRAMPA_ENABLE_DEBUG engrampa```

Issue #388, opened 2020-06-04
Suggestion: Remove final dialogbox


liar666:
#### Expected behaviour
No dialogbox displayed. Eventually change the (color of) some GUI element to indicate that decompression is terminated OK/Not OK, but do not create a new window/dialog.

#### Actual behaviour
At the end of any decompression, a dialogbox is shown, displaying "Extraction completed successfully" asking if the user wants to Quit/Show the files/Show the files and Quit/Close.
The opening of this dialogbox is quite annoying, particularly when decompression takes time, and you switch to another task in the meantime, and then the "termination" pops up and takes the focus.

#### Steps to reproduce the behaviour
- Run 
  $ engrampa <archive>
- Select files/dirs inside de archive
- Click "Extract" button
- Select a destination dir
- Hit "Extract" button

#### MATE general version
Engrampa Archive Manager 1.24.0

#### Package version
Engrampa Archive Manager 1.24.0

#### Linux Distribution
XUbuntu 20.04

dancer69:
I'm using engrampa because of this future mostly, which I found very useful in some situations. So, I disagree. I'd like this future to stay. It could be added a configuration option to enable/disable the dialogbox though.

PR #474, opened 2023-03-24
Add the function of automatically closing the dialog box after decompression

zhuyaliang:
Add the function of automatically closing dialog after successful decompression
#388 
[2023年03月24日 屏幕视频 11时40分17秒.webm](https://user-images.githubusercontent.com/36913152/227419075-367636cb-c785-448a-aea5-5c07d1f636f5.webm)

### DIFF
--- a/data/org.mate.engrampa.gschema.xml.in
+++ b/data/org.mate.engrampa.gschema.xml.in
@@ -150,6 +150,10 @@
       <default>true</default>
       <summary>Recreate the folders stored in the archive</summary>
     </key>
+    <key name="close-dialog" type="b">
+      <default>false</default>
+      <summary>Close dialog after decompression</summary>
+    </key>
   </schema>
   <schema gettext-domain="@GETTEXT_PACKAGE@" id="org.mate.engrampa.dialogs.add" path="/org/mate/engrampa/dialogs/add/">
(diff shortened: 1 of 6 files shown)


lukefromdc (commented):
I found some inconsistancy in applying the new preference. If the extract dialog is open and dconf-editor is also open w this preference showing, toggling it in deconf-editor also toggles it in the dialog, but toggling it in the dialog does NOT show up in dconf-editor.

Also, to apply a change of this preference requires restarting engrampa seemingly no matter which way it is applied

zhuyaliang:
@lukefromdc This is a normal phenomenon. I deliberately designed it this way. ```deconf editor``` plays a global role and can change the ```close_dialog_checkbutton``` toggles status. toggles button only works in this decompression, and it does not affect the deconf editor

If you want to permanently close the dialog box after decompression, please change gsettings. If you just want to close the dialog box after the current decompression is completed, please use the toggle button

lukefromdc (approved):
OK, this works on opening engrampa and toggling the box to close or not close the dialog. Looks like the only time I run into an issue (other than the differmet dconf-editor behaviior) is toggling this between multiple extractions from the same instance of the dialog or possibly the same instance of engrampa, a less common use case

PR #476, opened 2023-04-14
action: Delete unused sort_by_entries code

zhuyaliang:
```sort_by_entries``` Not appearing in any UI file, users cannot use these sorting options, so this is useless code


raveit65 (approved):
LGTM

lukefromdc (approved):
Builds and runs fine, a quick test showed no UI difference whatsoever, as expected

PR #477, opened 2023-04-14
file-utils: Include <strings.h> for strcasecmp

fweimer-rh:
The _XOPEN_SOURCE macro definition overrides _DEFAULT_SOURCE and disables the declaration in <string.h>.

This avoids an implicit function declaration and build failures with future compilers.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/src/file-utils.c
+++ b/src/file-utils.c
@@ -27,6 +27,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
 #include <ctype.h>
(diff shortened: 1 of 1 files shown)


raveit65 (approved):
LGTM

lukefromdc (approved):
Builds and runs fine, very simple change. No apparent difference in size of compiled binary down to 1/10 kb resolution

PR #479, opened 2023-05-08
Add MIME aliases for gzip and bzip2

gber:
The application/gzip alias is e.g. specified by FileRoller and Ark, application/bzip2 is specified by FileRoller.

If another archiver is installed these aliases seem to be preferred and it is not possible to override the association in a `mimeapps.list` file if engrapa's desktop file doesn't provide them.


zhuyaliang (approved):
Agree to add a new mimetype

lukefromdc (approved):
On my long-established system (nothing like a new empty user of course) on Debian Unstable I didn't notice any behavioral change from this: I have file-roller installed and never had a problem with engrampa not being selectable for the application to open .gz files with. No idea what the default was after so many years though.

lukefromdc:
Called for another review if possible as I simply don't know much about this

cwendling (approved):
You could add [`application/x-bzip2`](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in#L1605) that is missing as well, but otherwise looks good.

gber:
> You could add [`application/x-bzip2`](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in#L1605) that is missing as well, but otherwise looks good.

Added.

PR #480, opened 2023-05-10
Fix failure to extract multiple files

### LINKED ISSUE #371: Bulk Extraction Failing (Right Click)
- OS: Kali Linux 2020.1 (x64)
- Engrampa: v1.24.0-2 [Debian Package]
- Xfce v4.14 [Debian Package]
- Thunar v1.8.12 [Debian Package]
- thunar-archive-plugin v0.4.0-2 [Debian Package]

- - -

So the issue is, upon selection two or more files to extract I can't extract them both. It will only do the first one, and stop.

Will generate some random tar.gz files

```
kali@kali:~/Downloads$ ls
kali@kali:~/Downloads$
kali@kali:~/Downloads$ for x in $( seq 1 3 ); do
>   dd if=/dev/urandom of=file${x}.bin bs=1M count=1
>   tar -zvcf file${x}.tar.gz file${x}.bin
>   rm -f file${x}.bin
> done
1+0 records in
1+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.0198063 s, 52.9 MB/s
file1.bin
1+0 records in
1+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.0192154 s, 54.6 MB/s
file2.bin
1+0 records in
1+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.0194023 s, 54.0 MB/s
file3.bin
kali@kali:~/Downloads$
kali@kali:~/Downloads$ file *
file1.tar.gz: gzip compressed data, from Unix, original size modulo 2^32 1054720
file2.tar.gz: gzip compressed data, from Unix, original size modulo 2^32 1054720
file3.tar.gz: gzip compressed data, from Unix, original size modulo 2^32 1054720
kali@kali:~/Downloads$
```
Now select them all, right click and then...

- selecting "Extract here" works as expected, all three extract as expected
- But doing "Extract To" just does the first one.

zhuyaliang:
[05-13-2023Screen video05^%41^%24 PM.webm](https://github.com/mate-desktop/engrampa/assets/36913152/c2b8e2cc-fe05-4fe0-9d04-0b64d1975ed9)
Fix failure to extract multiple files
Fix #371 

## test
```
mkdir /tmp/test
cd /tmp/test
touch test{1,2,3}.file
find -type f -name test"*"file -execdir tar -zcvf '{}.tar.gz' '{}' \; | xargs rm -rf
engrampa -f *.gz
```

### DIFF
--- a/src/glib-utils.c
+++ b/src/glib-utils.c
@@ -510,6 +510,9 @@ _g_path_get_base_name (const char *path,
 	if (junk_paths)
 		return _g_path_get_file_name (path);
 
+	if (base_dir == NULL)
+		return (path[0] == '/') ? path + 1 : path;
+
 	base_dir_len = strlen (base_dir);
 	if (strlen (path) < base_dir_len)
 		return NULL;


lukefromdc (commented):
Another fix for an issue I don't know how to invoke, but again it builds fine and on extracting multiple files from the GUI using the extract menubar item I didn't have any problems. That works from git master too though, so we need a test by someone who can invoke the issue

zhuyaliang:
@lukefromdc Need to use ```Extract To``` test

lukefromdc (approved):
OK, found it thanks to the video: "extract to" comes up in caja-engrampa (caja's right-click menu). On master as of prior to this, only one file is extracted but with this installed all of the files are extracted. Simple change and it works.

PR #482, opened 2023-05-11
Uniform file chooser dialog margin size

### LINKED ISSUE #184: "Archive > New" window, unneeded margins.
#### Expected behaviour
Engrampa's file chooser does not match native gtk file chooser.

#### Actual behaviour
Window for creating new archive uses margins. It should not use margins to match gtk file chooser appearance. 

Just go to menubar, Archive -> New. 

See screenshot. Engrampa's "New" window on the left side, standard file chooser on the right side.
https://i.imgur.com/FvLsMbQ.png

#### Package version
engrampa 1.18.2-1

#### Linux Distribution
Manjaro

zhuyaliang:
The file chooser dialog margin size should remain the same， Open、New、Extract
Fix #184

### DIFF
--- a/src/ui/new.ui
+++ b/src/ui/new.ui
@@ -213,7 +213,6 @@
   </object>
   <object class="GtkFileChooserDialog" id="dialog">
     <property name="can_focus">False</property>
-    <property name="border_width">5</property>
     <property name="modal">True</property>
     <property name="window_position">center-on-parent</property>
     <property name="type_hint">normal</property>


lukefromdc (approved):
I never noticed an issue with this but don't have an eye for subtle differences between similar dialogs. Didn't get any problems with this built and installed however. Change is very simple

raveit65:
I do not get why it is bad to use a margin.
It is simple a question of design.
Personal i won't remove the margin. But ignore my noise.
The world won't be better with or without a margin :)

raveit65 (approved):
LGTM,
because complete unimportant.

PR #483, opened 2023-05-12
Set the current folder of the file chooser dialog using a local file name

### LINKED ISSUE #134: The folder contents could not be displayed - engrampa with --default-dir and --extract
in a terminal:

`engrampa --default-dir=./anyfolder/ --extract filecompressed
`

bugfolderengrampa

zhuyaliang:
Fix #134

### DIFF
--- a/src/dlg-extract.c
+++ b/src/dlg-extract.c
@@ -295,6 +295,7 @@ dlg_extract__common (FrWindow *window,
 {
 	DialogData *data;
 	GtkWidget  *button;
+	const char *extract_default_dir;
 
 	data = g_new0 (DialogData, 1);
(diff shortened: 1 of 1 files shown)


lukefromdc:
This has literally never been part of my workflow so I can't really wrap my head around the test

lukefromdc (commented):
Could not wrap my head around the test as I do not use engrampa from the command line in this way, but I can confirm this builds and didn't break extracting files from an archive from the UI. That however is all I can confirm so this is a comment

zhuyaliang:
@raveit65 ```Filecompressed``` should be a real compressed file
2023-05-15 10-15-53屏幕截图

raveit65 (commented):
LGTM,
extracting a compressed file into an existing directory works fine.

raveit65 (approved):

PR #485, opened 2023-05-15
Add "Extract to subdirectory" option to the extract dialog

### LINKED ISSUE #449: Option to always extract to sub folder.
#### Expected behaviour
A setting that makes every archive extract into PATH/archive_name/

#### Actual behaviour
Can't even on-demand request the application to extract into a subdir, populating the directory name automatically.

Extract any archive.

#### MATE general version
N/A

#### Package version
Engrampa 1.26.0

#### Linux Distribution
Manjaro XFCE

#### Link to bugreport of your Distribution (requirement)
N/A

#### Why do I want this?
Many archives are a flat structure, dumping their contents directly into the Downloads folder is far from ideal.
If it could automatically extract into sub folders it is much easier to have things organized.

xmusjackson:
This PR aims to add an "Extract to subdirectory" option to the extract dialog which allows users to extract the contents of an archive into a subdirectory of the archive's name, prompting for creation of the directory if necessary.

This should resolve #449


src/dlg-extract.c
```diff
@@ -237,6 +247,7 @@ extract_cb (GtkWidget   *w,
 
 	path_list_free (file_list);
 	g_free (extract_to_dir);
+	g_free (sub_dir);
```
zhuyaliang, line 250: Should be:
```
if (sub_dir != NULL)
    g_free (sub_dir);
```

src/file-utils.c
```diff
@@ -624,6 +624,22 @@ path_list_dup (GList *path_list)
 	return g_list_reverse (new_list);
 }
 
+/*Concatenates sub_dir and extract_to_dir
+ returns extract_to_dir + "/" + sub_dir */
+char *
+concat_sub_dir (const char* extract_to_dir, const char* sub_dir) {
+    int total_size = strlen(extract_to_dir) + strlen(sub_dir) + 2;
+	char *return_str;
+	char *temp_string = g_strconcat(extract_to_dir,
```
zhuyaliang, line 633: Code indentation, alignment, parenthesis formatting

src/file-utils.h
```diff
@@ -73,6 +73,9 @@ const char*         get_file_mime_type           (const char  *uri,
                     				  gboolean     fast_file_type);
 const char*         get_file_mime_type_for_path  (const char  *filename,
                     				  gboolean     fast_file_type);
+
```
zhuyaliang, line 76: Extra spaces

src/preferences.h
```diff
@@ -63,6 +63,7 @@
 #define PREF_EXTRACT_SKIP_NEWER		"skip-newer"
 #define PREF_EXTRACT_CLOSE_DIALOG	"close-dialog"
 #define PREF_EXTRACT_RECREATE_FOLDERS	"recreate-folders"
+#define PREF_EXTRACT_CREATE_SUBDIR  "create-subdirectory"
```
zhuyaliang, line 66: Maintain code alignment

src/dlg-extract.c
xmusjackson: Fixed in latest commit.

src/file-utils.c
xmusjackson: I missed the space between the function calls and the parameters here as I'm not used to that convention, but the spacing issues here and in the following conversations were an issue of OSS code handling spacing automatically. I attempted to fixed the issues in leafpad.

src/file-utils.h
xmusjackson: Mistaken spacing here, I forgot to fix this before the initial commits. Fixed in latest

src/file-utils.c
```diff
@@ -624,6 +624,24 @@ path_list_dup (GList *path_list)
 	return g_list_reverse (new_list);
 }
 
+/*Concatenates sub_dir and extract_to_dir
+ returns extract_to_dir + "/" + sub_dir */
+char *
+concat_sub_dir (const char* extract_to_dir, const char* sub_dir)
+{
+	int total_size = strlen (extract_to_dir) + strlen (sub_dir) + 2;
```
zhuyaliang, line 632: unused variable ```total_size```

src/fr-window.c
```diff
@@ -6321,8 +6324,12 @@ extract_data_new (GList       *file_list,
 
 	edata = g_new0 (ExtractData, 1);
 	edata->file_list = path_list_dup (file_list);
-	if (extract_to_dir != NULL)
+	if (sub_dir != NULL)
+		edata->sub_dir = sub_dir;
```
zhuyaliang, line 6328: assignment discards ‘const’ qualifier from pointer target type

src/dlg-extract.c
```diff
@@ -184,11 +186,13 @@ extract_cb (GtkWidget   *w,
 	overwrite = gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (GET_WIDGET ("overwrite_checkbutton")));
 	skip_newer = ! gtk_toggle_button_get_inconsistent (GTK_TOGGLE_BUTTON (GET_WIDGET ("not_newer_checkbutton"))) && gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (GET_WIDGET ("not_newer_checkbutton")));
 	junk_paths = ! gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (GET_WIDGET ("recreate_dir_checkbutton")));
+	extract_sub_dir = gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (GET_WIDGET ("create_subdir_checkbutton")));
```
zhuyaliang, line 189: Should be added ```fr_window_set_extract_default_dir (window, extract_to_dir, !extract_sub_dir);```, otherwise there may be issues when selecting to extract multiple compressed files

src/file-utils.c
```diff
@@ -624,6 +624,24 @@ path_list_dup (GList *path_list)
...
+	char *return_str;
+	char *temp_string = g_strconcat (extract_to_dir,
```
zhuyaliang, line 634: Why use temporary variables here? Can I directly use the ```returen_str``` variable

src/dlg-extract.c
```diff
@@ -220,6 +224,11 @@ extract_cb (GtkWidget   *w,
 	else
 		base_dir = NULL;
 
+	if (extract_sub_dir)
+		sub_dir = remove_extension_from_path (g_filename_display_basename (fr_window_get_archive_uri (window)));
```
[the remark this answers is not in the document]
xmusjackson: I was thinking about how to implement this. This function (remove_extension_from_path) could be re-implemented to truncate one or two extension names from the end of the file, but I'd be worried about files which have dots in the name already, such as "some_program_1.26.zip". Perhaps an array of the supported filetypes could be used for this, and anything not matching a supported type will remain in the name. Do you or anyone have any suggestions on how this should be handled while following conventions?

src/file-utils.c
xmusjackson: I do this in order to free the results of g_strconcat inside the function so that it doesn't need to be freed later wherever it's called (though this could be done if preferred this way). return_str is then assigned the results of a g_strdup_inline on temp_string so that temp_string can be freed. If there is a better way or a way that follows conventions more closely, I could certainly change this.

zhuyaliang: My suggestion is to remove this function and use ```g_build_filename```
```
    if (sub_dir != NULL)
        edata->sub_dir = sub_dir;
    if (extract_to_dir != NULL && sub_dir == NULL) {
        edata->extract_to_dir = g_strdup (extract_to_dir);
    } else if (extract_to_dir != NULL && sub_dir != NULL)
        edata->extract_to_dir = g_build_filename (extract_to_dir, sub_dir, NULL);
```

src/dlg-extract.c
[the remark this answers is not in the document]
zhuyaliang: Perhaps you can get inspiration from  [FrExtensionType](https://github.com/mate-desktop/engrampa/blob/master/src/fr-init.c#L119)

src/dlg-extract.c
xmusjackson: I think I understand this function, my latest commit should solve this. The commit message conveys my current understanding of this functionality, but I haven't been through the code to fully grasp how non-interactive extraction is implemented. If I've misunderstood something, please let me know.

src/fr-window.c
xmusjackson: I suppose I didn't notice I'd forgotten to use g_strdup since sub_dir was never modified (sub_dir is really only a member of ExtractData for later implementations, such as a command line option or the like). Fixed in latest commits.

src/file-utils.c
xmusjackson: Ah, whoops. Mistakenly left this after reimplementing concat_sub_dir. Fixed in latest commits.

src/fr-window.c
```diff
@@ -6321,8 +6324,12 @@ extract_data_new (GList       *file_list,
 
 	edata = g_new0 (ExtractData, 1);
 	edata->file_list = path_list_dup (file_list);
-	if (extract_to_dir != NULL)
+	if (sub_dir != NULL)
+		edata->sub_dir = g_strdup (sub_dir);
```
zhuyaliang, line 6328: Don't forget to release memory at the end

zhuyaliang: It would be even better if you could check if the sub_dir is NULL by the way

xmusjackson: @mbkma This was my understanding as well, but I figured there might be a reason still
@zhuyaliang the code shows cases where extract_data_free is called on an ExtractData with a null base_dir member, but it's never checked and g_free is called on it either way. In fact, within dlg-extract.c, base_dir is freed immediately after calling extract_data_new(), regardless of whether it initialized to NULL.

zhuyaliang: thank you

src/dlg-extract.c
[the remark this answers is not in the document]
zhuyaliang: So currently, there's only one issue left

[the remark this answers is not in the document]
xmusjackson: I'm working on the implementation, but I wanted to get some feedback on it before I commit anything. A basic test is as follow: 
`#include <stdio.h>
#include <string.h>

#include <glib.h>
#include <gio/gio.h>

const char* file_ext[] = {
	 ".7z",
	 ".a",
	 ".ace",
	 ".alz",
	 ".ar",
	 ".arc",
	 ".arj",
	 ".bin",
	 ".br",
	 ".bz",
	 ".bz2",
	 ".cab",
	 ".cbr",
	 ".cbz",
	 ".cpio",
	 ".deb",
	 ".ear",
	 ".epub",
	 ".exe",
	 ".gz",
	 ".iso",
	 ".jar",
	 ".lha",
	 ".lrz",
	 ".lzh",
	 ".lz",
	 ".lzma",
	 ".lzo",
	 ".odp",
	 ".ods",
	 ".odt",
	 ".otp",
	 ".ots",
	 ".ott",
	 ".rar",
	 ".rpm",
	 ".rz",
	 ".sit",
	 ".swm",
	 ".tar",
	 ".tar.7z",
	 ".tar.br",
	 ".tar.bz",
	 ".tar.bz2",
	 ".tar.gz",
	 ".tar.lrz",
	 ".tar.lz",
	 ".tar.lzma",
	 ".tar.lzo",
	 ".tar.xz",
	 ".tar.Z",
	 ".tar.zst",
	 ".taz",
	 ".tbz",
	 ".tbz2",
	 ".tgz",
	 ".tlz",
	 ".txz",
	 ".tzma",
	 ".tzo",
	 ".udeb",
	 ".war",
	 ".wim",
	 ".xz",
	 ".z",
	 ".Z",
	 ".zip",
	 ".zoo",
	 ".zst",
	NULL
};

gchar *
remove_extension_from_path (const gchar* path)
{
	int         len;
	int         p;
	int         final_p;
	const char *ptr = path;
	char       *new_path;

	if (! path)
		return NULL;

	len = strlen (path);
	if (len == 1)
		return g_strdup (path);

	p = len - 1;
	final_p = p;
	for (int j = 0; ((p > 0) && (j < 2)); p--) {
		if (ptr[p] == '.') {
			for (int i = 0; (file_ext[i] != NULL); i++) {
				if (! (g_strcmp0 (path + p, file_ext[i]))) {
					final_p = p;
					len = p;
				}
			}
			j++;
		}
	}

	if (p == 0)
		final_p = len;
	new_path = g_strndup (path, (guint) final_p);

	return new_path;
}

int main() {
	char* filename1 = "archive_1.26.zip";
	char* filename2 = "archive.org.tar.gz";
	char* filename3 = ".hidden.zip";
	char* filename4 = ".zip"; // in case there's such a file
	char* filename5 = "archive.cpio.zip"; // keeps first extension here, as 
										 // engrampa only opens the first

	char* trunc_filename1 = remove_extension_from_path(filename1);
	char* trunc_filename2 = remove_extension_from_path(filename2);
	char* trunc_filename3 = remove_extension_from_path(filename3);
	char* trunc_filename4 = remove_extension_from_path(filename4);
	char* trunc_filename5 = remove_extension_from_path(filename5);

	printf("Filename 1: %s, Truncated: %s\n", filename1, trunc_filename1);
	printf("Filename 1: %s, Truncated: %s\n", filename2, trunc_filename2);
	printf("Filename 1: %s, Truncated: %s\n", filename3, trunc_filename3);
	printf("Filename 1: %s, Truncated: %s\n", filename4, trunc_filename4);
	printf("Filename 1: %s, Truncated: %s\n", filename5, trunc_filename5);
}
`
My concern with this method is that it would add a lengthy array initialization to the beginning of file-utils.c, but this could be useful for other functions, like get_file_extension. I have an implementation written but not commited for this reason, however it appears to work as expected. The function "remove_extension_from_path()" above is exactly as it appears in the commit.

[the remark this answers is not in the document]
zhuyaliang: We should use existing functions to complete this work
```
gchar *
remove_extension_from_path (const gchar *path)
{
    const char *ext;

    if (path == NULL)
        return NULL;

    ext = get_archive_filename_extension (path);
    if (ext == NULL || strlen (ext) == strlen (path))
        return g_strdup (path);
    else
        return g_strndup (path, strlen (path) - strlen (ext));
}

```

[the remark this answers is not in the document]
xmusjackson: That seems a more elegant approach, thanks. Implemented in latest commit. 

Either way, there is one issue I've run into while testing this feature. If one attempts to extract an archive (i.e. archive.xyz.zip), into a directory (i.e. /home/user) where a file exists with the name of the new subdirectory (i.e. archive.xyz), the extraction will fail and draw the dialog box which shows command line output, displaying the message `"ERROR: Can not open output file : Not a directory: ..."` 
This shouldn't be a huge issue as it's a relatively unlikely situation, but it seems worth noting.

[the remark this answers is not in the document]
zhuyaliang: @xmusjackson 
Thank you for your testing. We should address this issue in this PR, and this is my suggestion
```
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -6321,6 +6321,7 @@ extract_data_new (GList       *file_list,
                  gboolean     ask_to_open_destination)
 {
        ExtractData *edata;
+       int i = 0;
 
        edata = g_new0 (ExtractData, 1);
        edata->file_list = path_list_dup (file_list);
@@ -6329,7 +6330,14 @@ extract_data_new (GList       *file_list,
        if (extract_to_dir != NULL && sub_dir == NULL) {
                edata->extract_to_dir = g_strdup (extract_to_dir);
        } else if (extract_to_dir != NULL && sub_dir != NULL)
+       {
                edata->extract_to_dir = g_build_filename (extract_to_dir, sub_dir, NULL);
+               while (g_file_query_exists (g_file_new_for_uri (edata->extract_to_dir), NULL))
+               {
+                       g_free (edata->extract_to_dir);
+                       edata->extract_to_dir = g_strdup_printf ("%s/%s(%d)", extract_to_dir, sub_dir, i++);
+               }
+       }
```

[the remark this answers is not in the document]
xmusjackson: Thanks, this works great. It might be advantageous though to use a hyphen or underscore instead of parentheses. The resulting sub-directory could then be used in bash and shell scripts without escaping characters.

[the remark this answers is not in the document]
zhuyaliang: >  It might be advantageous though to use a hyphen or underscore instead of parentheses.
You can freely design, this is just my suggestion， Please rebase the commit

zhuyaliang:
@xmusjackson Please rebase the commit

src/dlg-extract.c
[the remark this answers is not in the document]
xmusjackson: @zhuyaliang Do you mean rebase just the commit with the bad commit message or squash all of these? The most recent commit has been rebased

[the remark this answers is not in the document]
zhuyaliang: I would like to squash all commit for the convenience of rolling history. Thank you

xmusjackson:
Should the while loop in extract_data_new check if the queried file is a directory if the overwrite flag is set to yes? This would allow merging like the overwrite option does when not extracting to a subdirectory.  Simplified code could be as follows:
```
diff --git a/src/fr-window.c b/src/fr-window.c
index 83a1584..4c5f461 100644
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -6331,7 +6331,9 @@ extract_data_new (GList       *file_list,
                edata->extract_to_dir = g_strdup (extract_to_dir);
        } else if (extract_to_dir != NULL && sub_dir != NULL) {
                edata->extract_to_dir = g_build_filename (extract_to_dir, sub_dir, NULL);
-               while (g_file_query_exists (g_file_new_for_uri (edata->extract_to_dir), NULL)) {
+               while ((g_file_query_exists (g_file_new_for_uri (edata->extract_to_dir), NULL)) && 
+                      (!((g_file_query_file_type (g_file_new_for_uri (edata->extract_to_dir), G_FILE_QUERY_INFO_NONE, NULL) == G_FILE_TYPE_DIRECTORY)) ||
+                       overwrite == FR_OVERWRITE_NO)) {
                        g_free (edata->extract_to_dir);
                        edata->extract_to_dir = g_strdup_printf ("%s/%s_%d", extract_to_dir, sub_dir, i++);
                }
```
This may not be necessary though so I haven't commited anything.

zhuyaliang:
File type checking should be added, otherwise we will lose the overwrite function.
This is the reference code
```
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -6331,7 +6331,7 @@ extract_data_new (GList       *file_list,
                edata->extract_to_dir = g_strdup (extract_to_dir);
        } else if (extract_to_dir != NULL && sub_dir != NULL) {
                edata->extract_to_dir = g_build_filename (extract_to_dir, sub_dir, NULL);
-               while (g_file_query_exists (g_file_new_for_uri (edata->extract_to_dir), NULL)) {
+               while (uri_exists (edata->extract_to_dir) && uri_is_file (edata->extract_to_dir)) {
                        g_free (edata->extract_to_dir);
                        edata->extract_to_dir = g_strdup_printf ("%s/%s_%d", extract_to_dir, sub_dir, i++);
                }
```
Thank you again for your work

zhuyaliang (approved):

zhuyaliang:
@mbkma Do you have any other suggestions

xmusjackson:
I've discovered another bug while testing. If a tar compressed with a format other than an expected format is opened to view or extract the tar file itself, (i.e. archive.tar.zip) and "Extract to subdirectory" is checked, a directory with the name "archive.tar.zip" is created as the subdirectory for extraction (suffixed if extracted to the same directory, of course). This is because while "get_file_extension" returns ".tar.zip" for this file, get_archive_filename_extension returns NULL as ".tar.zip" doesn't match any of the supported file types in file_ext_type[]. This could be mitigated by rechecking files with extensions that begin with ".tar" but don't match expected compressed tar extension names (as these would be opened as just the last extension's filetype anyway):
```
@@ -487,6 +487,15 @@ get_archive_filename_extension (const char *filename)
                        return ext;
        }
 
+       if (strncasecmp (ext, ".tar", 4) == 0) {
+               for (i = G_N_ELEMENTS (file_ext_type) - 1; i >= 0; i--) {
+                       if (file_ext_type[i].ext == NULL)
+                               continue;
+                       if (strcasecmp (ext + 4, file_ext_type[i].ext) == 0)
+                               return ext + 4;
+               }
+       }
+
        return NULL;
 }
 
```

zhuyaliang:
@xmusjackson I think it would be better to fix it in the ```get_file_extension``` function
This is my fix code
```
--- a/src/file-utils.c
+++ b/src/file-utils.c
@@ -507,6 +507,7 @@ get_file_extension (const char *filename)
        int         len;
        int         p;
        const char *ext;
+       const char *tar_exts[] = {".7z", ".br", ".bz", ".bz2", ".gz", ".lrz", ".lz", ".lzma", ".lzo", ".xz", ".Z", ".xst", NULL};
 
        if (filename == NULL)
                return NULL;
@@ -522,10 +523,18 @@ get_file_extension (const char *filename)
                return NULL;
 
        ext = filename + p;
+       p = 0;
        if (ext - 4 > filename) {
                const char *test = ext - 4;
                if (strncmp (test, ".tar", 4) == 0)
-                       ext = ext - 4;
+               {
+                       while (tar_exts[p] != NULL)
+                       {
+                               if (strcmp (ext, tar_exts[p]) == 0)
+                                       ext = ext - 4;
+                               p++;
+                       }
+               }
        }
        return ext;
 }
```

raveit65:
Can you please use a commit header and a commit body?
Use a blank line between both.
For example:
```
Add "Extract to subdirectory" option         --> commit header

This commit adds an                                    -->  first line of commit body
"Extract to subdirectory" option to the extract dialog which allows the user
to extract the contents of the archive to a directory with the name of the
archive (without the extension.) The user will be prompted if the directory
must be created.

etc.
```
This looks better in terminal with the `git log` command and in release notes.
Use the `git rebase -i ........` command for editing and choose the rewrite option.

zhuyaliang:
@raveit65 Do you have any further suggestions？

raveit65 (approved):
LGTM,
does what it promised.

PR #486, opened 2023-05-15
Add correct jar mime type

noeppi-noeppi:
According to https://www.iana.org/assignments/media-types/application/java-archive, the MIME type for jar files is `application/java-archive`. However, currently, engrampa only lists `application/x-java-archive` in its desktop file. This pull request adds the `application/java-archive` MIME type as well as the other two deprectaed aliases listed in the document above (`application/jar` and `application/jar-archive`) to the list.


zhuyaliang:
@cwendling @lukefromdc Do we need to accept this PR? I am not familiar with jar type files

cwendling (approved):
I don't use that either, but as we already have `application/x-java-archive`, it makes sense to accept these as well.

According to [IANA](https://www.iana.org/assignments/media-types/application/java-archive):

 > It is recommended that applications which use a media type/media subtype other than "application/java-archive", including "application/jar", "application/jar-archive", "application/x-java-archive" should be updated to use "application/java-archive".

suggesting they all are aliases, even though [shared-mime-info](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in#L2530) doesn't know about `application/jar` or `application/jar-archive` yet (but does know `application/x-jar`).

---

This makes me wonder whether we shouldn't find a way to build these MIME lists automatically somehow, like querying all aliases and subtypes from the MIME database when building this… but that's not an idea for this PR, although it might be worth pursuing if anybody's interested.

raveit65 (approved):
LGTM,
small change.....

PR #487, opened 2023-05-19
Add meson compilation support

zhuyaliang:
## test
```
meson setup _build --prefix=/usr

ninja -C _build

meson install -C _build
```


lukefromdc (commented):
I got this error at the start of the build after an apparent successful configuration:
```

ninja: Entering directory `_build'
ninja: error: '../data/engrampa.appdata.xml.in', needed by 'data/engrampa.appdata.xml', missing and no known rule to make it

```

This is on Debian Unstable updated a few days ago

zhuyaliang:
@lukefromdc Fixed

lukefromdc (approved):
This now works, configure, build, install, and uninstall all work and uninstall claims not to leave any remnant files. Question: how do we enable libmagic for meson builds. I use meson but don't know much about setting up nondefault options with it

zhuyaliang:
@lukefromdc Set default options in the ```meson_options.txt``` file,Use the ```-D``` option to change the default value
```
[test@localhost engrampa]$ cat meson_options.txt 
option('run-in-place', 
  type : 'boolean', 
  value : false, 
  description : 'Load ui data from the source tree'
)

option('caja-actions',
  type : 'boolean', 
  value : true, 
  description : 'Build the nautilus context menu actions'
)

option('packagekit',
  type : 'boolean', 
  value : true, 
  description : 'Build PackageKit installer for missing utilities'
)

option('magic',
  type : 'boolean', 
  value : true, 
  description : 'Use libmagic to detect file type'
)

option('cpio',
  type : 'string',
  value : 'auto',
  description : 'Path to the cpio program'
)

meson setup _build --prefix=/usr -Dmagic=false
```

lukefromdc:
This may or may not be a bug, but meson refused to install over an existing install of engrampa, yielding this error:
`
ERROR: Destination '/usr/share/help/bg/engrampa/figures/engrampa_main_window.png' already exists and is not a symlink`

lukefromdc:
Using `meson setup _build --prefix=/usr -Dmagic=true` and removing my normal install prior to installation gave me engrampa with working libmagic support, thanks

zhuyaliang:
> already exists and is not a symlink

This is not a bug [eom](https://github.com/mate-desktop/eom/pull/339)

lukefromdc:
I got this warning but another successful build w meson, looks like it re-ran pulling mate-submodules. Didn't notice that in my last test, only saw this when building the revised git master for a Debian package. Not sure if this is even a problem, that depends on the effect on offline builds:

```
Run-time dependency mate-submodules found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency mate-submodules

Executing subproject mate-submodules 
```

Issue #206, opened 2018-07-05
Cannot extract tar.gz


smiddy:
#### Expected behaviour
Having an archive archive.tar.gz and extract it.

#### Actual behaviour
I get the error:

tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors

#### Steps to reproduce the behaviour
The error does not occur on all tar.gz. files. E.g., when I create an archive with engrampa, I can extract it afterwards.
The archive itself is o.k, I extracted it on a CentOS 7 mashine with a different program.

#### MATE general version
1.20

#### Package version
1.20

#### Linux Distribution
Xubuntu 18.04 64bit

cranes-bill:
So the file probably really is not a tar format.

smiddy:
Unfortunately, there is some sensitive data in the archive. I think i produced it once on my CentOs mashine as well.

@vkareh should I try to produce a tar.gz archive which resultstin the error?

vkareh:
@smiddy, yes please. That would be very helpful. Any file that fails like that would work for me, just make sure that it only fails in engrampa (file-roller and `tar xvfz` should be able to open it normally) so that I can reproduce the issue. Thanks!

cranes-bill:
This error occurs when the file is not the same format as extension supposed. Just check it with `file`
![capture du 2018-08-02 15 27 14](https://user-images.githubusercontent.com/5161278/43583906-9117afc2-9669-11e8-8635-712e43f7e83d.png)

The error should be reproducible with the attached file!
[README.tar.gz](https://github.com/mate-desktop/engrampa/files/2253467/README.tar.gz)

lukefromdc:
Tar cannot open this file, I got

```
luke@ubuntu:~/Desktop$ tar -xf '/home/luke/Desktop/README' 
tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors

```
So either the file is bad(most likely) or there is a problem in tar itself. I don't know if there is some new kind of tar archive that is not supported by even the tar version shipped now in Debian Unstable, or if there are some random problems where the problem archives are first being made.

vkareh:
I would say that if `tar` cannot open the file, it's probably beyond us to fix it... (just tried it and get the same issue - no matter which flags I pass on to `tar`, it just doesn't work, so this file is corrupted).

@cranes-bill - how did you create this file?

cranes-bill:
[...]

PR #491, opened 2023-06-14
Fix MIME detection logic from #490

cwendling:
A blooper has been made there:

* if ENABLE_MIME is set, the intention was to try, in order: magic, content, filename; but it was made filename, content, magic (which was the same as before the changes);
* if ENABLE_MIME is not set, the intention was to try, in order: filename, content, magic; but it has been made magic, content, magic (notice the duplicate, and the missing "filename").

This probably doesn't change much in the wild as magic is gonna work most of the time, but it's especially problematic that the non-libmagic case doesn't have the filename test.

Anyway, fix this so the code is consistent, and we retain the behavior for the non-libmagic case, and have the new expected one for the libmagic case.

CC @sinha-toyeesh

### DIFF
--- a/src/fr-archive.c
+++ b/src/fr-archive.c
@@ -1102,16 +1102,22 @@ load_local_archive (FrArchive  *archive,
 
 	old_command = archive->command;
 
+	/* With libmagic, give priority to magic match, and fall back on filename
+	 * match.  Do it the other way around if we don't have libmagic. */
 #if ENABLE_MAGIC
- 	mime_type = get_mime_type_from_filename (archive->local_copy);
-#else
 	mime_type = get_mime_type_from_magic_numbers (archive->local_copy);
+#else
+	mime_type = get_mime_type_from_filename (archive->local_copy);
 #endif
 
 	if (! create_command_to_load_archive (archive, mime_type)) {
 		mime_type = get_mime_type_from_content (archive->local_copy);
 		if (! create_command_to_load_archive (archive, mime_type)) {
+#if ENABLE_MAGIC
+			mime_type = get_mime_type_from_filename (archive->local_copy);
+#else
 			mime_type = get_mime_type_from_magic_numbers (archive->local_copy);
+#endif
 			if (! create_command_to_load_archive (archive, mime_type)) {
 				archive->command = old_command;
 				archive->content_type = mime_type;


sinha-toyeesh:
Oh I get it now, thanks for the clarification, appreciate the thorough explanation @cwendling, you've been a great help.

lukefromdc (approved):
Worked fine here on same test files with libmagic enabled

zhuyaliang (approved):
Safer check of mime type

lukefromdc:
Turns out we have a problem with tar.xz files at this point, the directory within is not being recognized starting with the PR this one was supposed to fix:
Before 8e33e604d87e90bda3cc748bc1627c164d72a16c
Engrampa_before_PRs

After 8e33e604d87e90bda3cc748bc1627c164d72a16c:

Engrampa_after_PRs

zhuyaliang:
Confirm that the problem exists

zhuyaliang:
Actually, this is also the correct handling method, but it only requires two processing steps
```
 [test@localhost test]$ file test.tar.xz 
test.tar.xz: XZ compressed data
[test@localhost test]$ xz -d test.tar.xz 
[test@localhost test]$ ls
test.tar
[test@localhost test]$ file test.tar 
test.tar: POSIX tar archive (GNU)
[test@localhost test]$ tar -xvf test.tar 
test/
test/main.c
test/b.c
test/main
test/libb.so
test/a.c
test/liba.so
[test@localhost test]$ 
```
@cwendling How do you view this issue? Do we revert back to the past？

cwendling:
Hum, I don't know… I assumed the intent of #490 was good, but maybe it isn't if libmagic doesn't tell us about those subtypes. Does the `g_content_type` thing work better?  If so, that could be the first (or only?) magic-aware handler (is it not just as capable as libmagic?).

cwendling:
@sinha-toyeesh what was the actual intent behind #490?  Did you have something you needed fixing, or was this for the same of making it seemingly more consistent?

Anyhow, this would require more testing but I'd personally try `g_content_type` first.  AFAICT it does both filename and magic matching, which is kind of the best of both world.  And it does the right thing for e.g. `.tar.xz`:

```console
$ gio info -a standard::content-type,standard::fast-content-type foobar.tar.xz
[…]
attributes :
  standard::content-type: application/x-xz-compressed-tar
  standard::fast-content-type: application/x-xz-compressed-tar
```

This said, I really don't know much about Engrampa's internals, nor history behind the filename/gcontenttype/libmagic trio, so I'm not necessarily the best one to give an opinion.

sinha-toyeesh:
I would say it say it was a bit of both as mentioned in the original quoted  #206, the solution to which, I assumed was an issue of inconsistent use of libmagic, however that was an incomplete picture, as you have elaborated in your example.

Issue #492, opened 2023-07-15
Engrampa seems to be broken after last git commits


L-U-T-i:
#### Expected behaviour
Open .tar.gz file in engrampa and see the directory inside of it (typically a package source file, which does contain a directory) 

#### Actual behaviour
Open .tar.gz file in engrampa but see only 1 file with the name of the directory

#### Steps to reproduce the behaviour
Open .tar.gz file with engrampa 1.27 latest git commit

#### MATE general version
1.27

#### Package version
1.27.0 - git20230615 - 79c9b8e, with pull request #486 patch applied

#### Linux Distribution
Rocky Linux 9.2

#### Link to bugreport of your Distribution (requirement)
no bugreport - it is a custom build

I suspect one of the commits from June 14 or 15 is the reason, as my previous build of commit from May 15, with pull requests #485 and #486, opens .tar.gz file correctly.

lukefromdc:
CONFIRMED-tar.gx files are broken, show only an unidentifiable file with the directory name, and that cannot be extracted. I had to roll back to my May 23 build a while back because I need to work with tar.gz files

zhuyaliang:
@lukefromdc Can you elaborate on the reproduction method so that I can solve this problem
![2023-07-17 10-10-40屏幕截图](https://github.com/mate-desktop/engrampa/assets/36913152/2fe5bb4d-a484-4d8c-add9-e6b12701e9b8)

lukefromdc:
On the desktop or a caja navigation window, select two or more files, then select "compress" in caja's right-click menu, and finally select "tar.gz" as the desired format. As a second test file, do the same with a single directory

It is possible you might need an older version of engrampa to make the files but I didn't test that and record results.

 Once you have this file, try to open it using engrampa from current git master. It will open, but in both cases you will have a single file of unknown type that cannot be extracted or opened. I even got this with a single-file tar.gz made with current git master engrampa

lukefromdc:
One exception: I was able to compress a single video file if and only if it had the correct extension on it to a tar.gz which would open with engrampa from git master

zhuyaliang:
@lukefromdc @L-U-T-i Please disable libmagic, I will resolve this issue later [491](https://github.com/mate-desktop/engrampa/pull/491)

lukefromdc:
I am using an older version for now, since this is a development release and not in use by distros we just need to fix or revert this before 1.28 goes out

lukefromdc:
[...]

PR #493, opened 2023-07-19
priority useing get_mime_type_from_filename to obtain mime type

zhuyaliang:
MAGIC returned incorrect mime type while processing tar compression type 
Fix [492](https://github.com/mate-desktop/engrampa/issues/492)

### DIFF
--- a/src/fr-archive.c
+++ b/src/fr-archive.c
@@ -1102,22 +1102,12 @@ load_local_archive (FrArchive  *archive,
 
 	old_command = archive->command;
 
-	/* With libmagic, give priority to magic match, and fall back on filename
-	 * match.  Do it the other way around if we don't have libmagic. */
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
Not perfect but a major improvement and usable. Will now open tar.gz files and can extract from them. If however you remove the extension from the filename, the previous behavior returns once the file is open. If libmagic is not enabled at all of course, the archive cannot be opened at all without the extension.

Good enough for me to use as my daily driver though

raveit65 (approved):
LGTM

lukefromdc (approved):
Approving, we need a fix and this works. We can worry about files without extension names inside a tar.gz later. As it stands now, engrampa is broken when libmagic is enabled so approving and will merge

PR #495, opened 2023-08-04
Check if the remaining disk space is sufficient when add extract

### LINKED ISSUE #344: full disk makes engrampa fail with meaningless error message
#### Expected behaviour

Engrampa to say, or at least log to the console, that the disk is full. Or print a stack trace. anything.

#### Actual behaviour

There doesn't seem to be any log output at all for errors, it just says "an error occured" or something. no stack trace. nothing. The console remains completely empty

create an archive on a full disk

#### MATE general version

I'm running xfce, not mate

#### Package version

1.22.2-1

#### Linux Distribution

manjaro

what is this

zhuyaliang:
Fix #344 
## test

```
[test@localhost engrampa]$ lsblk  /dev/sda1 
NAME MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda1   8:1    1 149.5M  0 part /mnt

[test@localhost engrampa]$ du -sh /tmp/mate-user-guide
190M	/tmp/mate-user-guide
[test@localhost engrampa]$ zip -r /tmp/mate-user-guide.zip /tmp/mate-user-guide/ -q
[test@localhost engrampa]$ du -sh /tmp/mate-user-guide.zip 
65M	/tmp/mate-user-guide.zip
[test@localhost engrampa]$ sudo  ./src/engrampa /tmp/mate-user-guide.zip -e /mnt/

```
2023-08-04 11-17-43屏幕截图

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -6419,6 +6419,34 @@ archive_is_encrypted (FrWindow *window,
 	return encrypted;
 }
 
+static gboolean
+archive_extract_check_disk_space_full (GPtrArray  *files,
+                                       const char *extract_to_dir)
+{
+	guint64 freespace;
+	guint64 archive_size = 0;
+	guint   i;
+
+	freespace = get_dest_free_space (g_filename_from_uri (extract_to_dir, NULL, NULL));
+
+	for (i = 0; i <files->len; i++)
+	{
+		FileData *fdata = g_ptr_array_index (files, i);
+		archive_size += fdata->size;
+	}
+	/* When the decompressed target file is less than 100MB,
+     2 * (target file size)  needs to disk space
+     and when it is greater than 100MB,
+     100MB + (target file size) needs to disk space */
+
+	if (archive_size <= 1024 * 1024 * 100)
+		archive_size = archive_size * 2;
+	else
+		archive_size += 1024 * 1024 * 100;
+
+	return freespace < archive_size * 2;
+}
+
 void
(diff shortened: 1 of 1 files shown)


L-U-T-i:
What about extracting a big archive with (partial) overwriting? In this case, less free space shall actually be needed...

lukefromdc:
Not sure what we should do with this given the effect on required disk space of the overwriting case-and the possiblity that the new files which overwrite old files are larger, or that some are larger and some are smaller.

That exact issue can come up in copying in Caja, where if I try to drag and drop a backup directory containing additional files to the directory it backs up, that destination is almost always half-full or more. Thus the copy job errors out with an insufficient disk space notice BEFORE the option to overwrite or skip can even be displayed.

raveit65 (approved):
LGTM

lukefromdc (approved):
This does work, and as engrampa now stands, an attempt to extract into a directory with insufficient space causes that directory to be entirely filled with partiall written files, and the extract operation erroring out when free space hits zero. Quite ugly. With this PR applied behavior is similar to Caja and won't attempt to overflow a directory.

We can worry about the overwrite case later it seems given the current situation

dtantsur:
Hi folks! There seems to be a regression here, although I don't see it obviously in the code. The check always assumes I'm extracting *the entire archive* or something like that. As a result, I cannot extract a 18MiB file into a disk with >3 GiB of free space. Opening the file in an editor works (and I'm using it as a workaround, but it's quite ugly).

Issue #496, opened 2023-09-14
Segfault on opening .deb files after recent Debian Unstable updates


lukefromdc:
#### Expected behaviour
.Deb files (Debian packages) open with engrampa

#### Actual behaviour
Engrampa closes immediately with a segfault on attempting to open a .deb file. Opening other types of archives (including RPM packages) works fine. No change in behavior if Engrampa is reconfigured and rebuilt.

#### Steps to reproduce the behaviour
On Debian Unstable updated after Sep 11 2023 attempt to open any .deb file

#### MATE general version
1.27 

#### Package version
1.27.1

#### Linux Distribution
Debian Unstable

#### Link to bugreport of your Distribution (requirement)
None as this is a local build

lukefromdc:
I got this backtrace running in gdb, segfault occurs on dropping a .deb file into the initially empty engrampa window

```
Starting program: /usr/bin/engrampa 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff31ff6c0 (LWP 217593)]
[New Thread 0x7fffea9fe6c0 (LWP 217594)]
[New Thread 0x7ffff29fe6c0 (LWP 217595)]
[New Thread 0x7ffff21fd6c0 (LWP 217596)]
[New Thread 0x7ffff0dff6c0 (LWP 217597)]
[Thread 0x7ffff0dff6c0 (LWP 217597) exited]
[New Thread 0x7ffff0dff6c0 (LWP 217598)]
[New Thread 0x7fffebfff6c0 (LWP 217599)]
[Thread 0x7ffff0dff6c0 (LWP 217598) exited]
[Thread 0x7fffebfff6c0 (LWP 217599) exited]
[New Thread 0x7fffebfff6c0 (LWP 217600)]
[New Thread 0x7ffff0dff6c0 (LWP 217601)]
[Thread 0x7fffebfff6c0 (LWP 217600) exited]
[Thread 0x7ffff0dff6c0 (LWP 217601) exited]
[New Thread 0x7ffff0dff6c0 (LWP 217602)]
[FR] ../src/fr-window.c:3983 (fr_window_drag_data_received):
	::DragDataReceived -->

[FR] ../src/fr-window.c:2774 (action_started):
	LOADING_ARCHIVE [START] (FR::Window)

[FR] ../src/fr-window.c:4133 (fr_window_drag_data_received):
[...]

PR #499, opened 2023-09-15
Fix Segfault on opening .deb files after recent Debian Unstable updates

zhuyaliang:
Fix https://github.com/mate-desktop/engrampa/issues/496


lukefromdc (approved):
Builds fine, works as intended. Can open .deb files in engrampa with dpkg 1.22.0 installed with this

lukefromdc:
What is the status of this? Debian files still cause engrampa to segfault with current git master

PR #507, opened 2024-01-23
7z: Fix accepting RAR archives when unar-open-zip=true

cwendling:
Offset was not properly updated when adding support for EPUB+ZIP in 84fb5cfe96263f79e50b68a5a6996e40c514cf74, leading to disabling both ZIP and RAR when the option unar-open-zip was enabled.

One can test with:
* One of the supported `7z` available (ideally with RAR support, but the error message itself is enough to see the difference)
* No (`un`)`rar` available (as it takes precedence)
* `unar` and `lsar` available (Debian package `unar`)
* Setting `gsettings set org.mate.engrampa.general unar-open-zip true`
* This patch (because if `unar` can take over it's tricky to tell):
  ```diff
  diff --git a/src/fr-command-unarchiver.c b/src/fr-command-unarchiver.c
  index c7aa5e6..0d3cf53 100644
  --- a/src/fr-command-unarchiver.c
  +++ b/src/fr-command-unarchiver.c
  @@ -246,7 +246,7 @@ fr_command_unarchiver_handle_error (FrCommand   *comm,
   
   const char *unarchiver_mime_type[] = { "application/zip",
   				       "application/x-cbr",
  -				       "application/x-rar",
  +				       //"application/x-rar",
   				       NULL };
   
   static const char **
  ```
* a RAR file (which is trickier than expected to make, so one could use something from [RAR-test-files](https://github.com/ssokolow/rar-test-files/tree/master/build), ideally one of the RAR3 ones for wider support)

Expected: the `org.mate.engrampa.general unar-open-zip` setting has no effect on RAR files.
Actual: It does have an effect.

This is probably actually pretty low impact as it seems the condition can only happen if another RAR-capable backend is available, but that's mostly by chance, and I didn't actually check whether `unar` works with RAR files if there's not (`un`)`rar` versions available.

### DIFF
--- a/src/fr-command-7z.c
+++ b/src/fr-command-7z.c
@@ -609,7 +609,7 @@ fr_command_7z_get_mime_types (FrCommand *comm)
 
 	if (g_settings_get_boolean (settings, "unar-open-zip") &&
 	    is_program_in_path ("unar") && is_program_in_path ("lsar"))
-		sevenz_mime_types [8] = NULL;
+		sevenz_mime_types [9] = NULL;
 	else
 		g_settings_set_boolean (settings, "unar-open-zip", FALSE);
 


zhuyaliang:
I don't use rar and cannot test this PR. Let others test it

lukefromdc:
I don't have rar installed either, and will never permit any software on my machines that asks for registration

cwendling:
@zhuyaliang I don't use RAR myself either normally, and don't have any means for creating a RAR archive.  If you're comfortable using the samples I linked (or any other you find) though, you don't actually need to be able to make any to test this.

@lukefromdc you don't have to, the only actual requirements for testing is a RAR archive, and one of the 7z implementations.  Even if it doesn't fully support RAR (through e.g. p7zip-rar or 7zip-rar), you'll get a different error message depending on whether out 7z wrapper claims support or not (it'll either tell you the format is not supported, or that you need to get extra software).

Whether or not you can get your hand on a RAR file you accept to feed engrampa, I don't know.  The samples I linked seemed legit enough for me to use them as a test, but I won't vouch for them either.

raveit65:
Hmm, i followed your instruction and compiled this branch with the proposed extra patch and i can open a rar archive with org.mate.engrampa.general unar-open-zip true/false.
But same i can do with engrampa from master branch too..... o_0
Only unar and 7zip from f38 repos is installed.
Do i miss something?
What is the expected test case?

cwendling:
@raveit65 are you positive you don't have an unrar executable?  Try also commenting the `"application/x-rar"` from all `src/fr-command*` files but 7z (there should be a match in `7z`, `unarchiver` and `rar`).  Otherwise, I'll have to check how it could possibly work in your case :)

cwendling:
@raveit65 OK I might have been unclear: to reproduce the bug on *master* you need to apply the patch, otherwise `unar` will take over opening the RAR archive.  Again, this has to be done on master to see the issue, not on this PR.

Basically the idea is that you should be able to apply this patch on a working version and not have *org.mate.engrampa.general unar-open-zip* affect opening RAR archives:
```diff
diff --git a/src/fr-command-rar.c b/src/fr-command-rar.c
index d6a724e..605ae55 100644
--- a/src/fr-command-rar.c
+++ b/src/fr-command-rar.c
@@ -671,7 +671,7 @@ fr_command_rar_handle_error (FrCommand   *comm,
 }
 
 const char *rar_mime_type[] = { "application/x-cbr",
-				"application/x-rar",
+				//"application/x-rar",
 				NULL };
 
 static const char **
diff --git a/src/fr-command-unarchiver.c b/src/fr-command-unarchiver.c
index c7aa5e6..0d3cf53 100644
--- a/src/fr-command-unarchiver.c
+++ b/src/fr-command-unarchiver.c
@@ -246,7 +246,7 @@ fr_command_unarchiver_handle_error (FrCommand   *comm,
 
 const char *unarchiver_mime_type[] = { "application/zip",
 				       "application/x-cbr",
-				       "application/x-rar",
+				       //"application/x-rar",
 				       NULL };
 
 static const char **
```

So:
1. Apply the above patch on master, and test opening a RAR archive with both values of *org.mate.engrampa.general.unar-open-zip*
2. Do the same on this PR

Expected is that master fails when *org.mate.engrampa.general.unar-open-zip=true*, and that this PR doesn't.

cwendling:
Adding @sc0w as a reviewer as he seem to have made some RAR changes, so maybe he can test a little easier :crossed_fingers:

raveit65:
WTF. A forgotten binary is installed.
```
[rave@mother engrampa-dist]$ rar

RAR 6.21   Copyright (c) 1993-2023 Alexander Roshal   16 Feb 2023
Trial version             Type 'rar -?' for help
```
I need to uninstall it clean and hope than i get better results.

cwendling:
@raveit65 you can also just apply the 2-line patch above so engrampa only has one RAR implementation so you know for sure it tries 7z, regardless of the binaries you have.

raveit65:
Seems my version of 7z don't like the test rar or doesn't support it. Listing fails.
```
[rave@mother engrampa-dist]$ 7z l /media/Data/Downloads/test-and-delete/testfile.rar3.cbr  
7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
p7zip Version 16.02 (locale=de_DE.UTF-8,Utf16=on,HugeFiles=on,64 bits,32 CPUs AMD Ryzen 7 5700G with Radeon Graphics          (A50F00),ASM,AES-NI)

Scanning the drive for archives:
1 file, 381 bytes (1 KiB)                       
Listing archive: /media/Data/Downloads/test-and-delete/testfile.rar3.cbr

ERROR: /media/Data/Downloads/test-and-delete/testfile.rar3.cbr : Can not open the file as archive
```
Sorry, no idea what can i do more to test your PR.



Errors: 1
[rave@mother engrampa-dist]
```

**Comment by raveit65** [comment]
Ok i read that p7zip is splitted in ubuntu. They have a package that called p7zip-rar.
I guess that p7zip-plugins from normal fedora repo doesn't ship proprietary bits like rar.

**Comment by cwendling** [comment]
> Now, `engrampa /media/Data/Downloads/test-and-delete/testfile.rar3.cbr` fails always to open the rar archive, regardless of whether `org.mate.engrampa.general unar-open-zip` is false or true.

On master?  Sure but is it the *same* error message?  The one in your capture normally means that engrampa thinks it has support but misses a package for it to work.  Basically, the MIME type would be on the list but misses a required external component (e.g. the actual command to run) -- which would be the case if you misses one of the required 7z components.
However, if you set `org.mate.engrampa.general.unar-open-zip=true` on master, you should get the *"Archive type not supported"* error instead, which means that there's no MIME mapping for that file (which is the bug :)).

This said, you should check with one of the `.rar` files: you tested a `.cbr`, which although is compressed with RAR has a different MIME type on recent-ish systems. That might be part of the problems you have reproducing :)

> From my side i trust you. Simply merge it :)

If you're tired enough of the shenanigans for testing this, sure I will, especially as although it has an impact, in won't for the large majority of users given how specific the setup is to trigger the bug in a way that actually affect usage :)

**Comment by raveit65** [comment]
> On master? Sure but is it the same error message? 

Yes, same error message from screenshot.

> This said, you should check with one of the .rar files: you tested a .cbr, which although is compressed with RAR has a different MIME type on recent-ish systems. That might be part of the problems you have reproducing :)

Ok, i will try again but my time contingent for this package is nearly running out

> If you're tired enough of the shenanigans for testing this, sure I will, especially as although it has an impact, in won't for the large majority of users given how specific the setup is to trigger the bug in a way that actually affect usage :)

I use engrampa like i deliver it for fedora and i have all dependencies installed to open all archives. I don't mind if a archive open with rar, 7zip, unar, unrar or what else. And everything works well :-)
Honestly, i don't understand people who say "i want to use only this special program" and create problems for themself and poor maintainers who work in there spare time. They should become a maintainer for themself.

After 11-12 years of maintaining MATE as upstream and downstream maintainer i am really tired to create special test environments on my box. In the end nobody of all this users say thank you for this work.
They only can file out bug reports in `one sentence` and without any information and ignore any help like eg. our bug reporting template.

And honestly there is one debian fork which do not have a own bug reporting system since years and their users can only file out reports here. I ignore such reports.
And yes, i am really tired of this bull shit.

**Comment by raveit65** [comment]
I tried again and now i get different results when opening  testfile.rar3.rar with engrampa from master branch.
With `unar-open-zip false` i get this error message.
```
Could not open this file type

There is no command installed for RAR archive files.
Do you want to search for a command to open this file?
```
With `unar-open-zip true` i get this error message
```
Could not open "testfile.rar3.rar"

Archive type not supported.
```
I hope this is the expected result.
Should i comment out `"application/x-rar"` when testing your branch?

raveit65:
With your branch i can open the archive with both gsettings cases.
Is this the expected behaviour?
I like to approve the PR.

cwendling:
> I hope this is the expected result.

Yes!!

> Should i comment out `"application/x-rar"` when testing your branch?
> Edit: after reading comments i should not :-)

Actually you should :)  You should basically do the same test, and get the "There is no command installed for RAR archive files." error with any value of the gsetting.

raveit65:
> Actually you should :) You should basically do the same test, and get the "There is no command installed for RAR archive files." error with any value of the gsetting.

I can confirm that PR brings back exactly this behavior with both gsettings values. :-)

raveit65 (approved):
LGTM

PR #512, opened 2024-02-24
 Distribute Meson build system

raveit65:
Installation with meson buildsystem works now from tarball created with `make dist`.
Btw. there is no meson.build file in /src/ui/ folder !
Not testet because i don't have any interests in using meson :P

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -31,6 +31,21 @@ EXTRA_DIST = 			\
 	README			\
 	autogen.sh
 
+# Distribute the Meson build system files as well
+EXTRA_DIST += \
+	meson.build			\
+	meson_options.txt		\
+	postinstall.py			\
+	caja/meson.build		\
+	data/meson.build		\
+	data/icons/meson.build		\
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
This works, and we can now complete meson builds BUT we have another issue: as previously stated, there is no meson.build file in src/ui AND in src/meson.build we only list
```

# Subdirectories

subdir('sh')
```
The build completes anyway, I've never used password protected archives etc, the menus and toolbars do work on meson builds so not sure what's going on there. Probably whoever did the meson port would know if this was left out intentionally or was an oversight.

Fixing it though can be an issue for another PR

raveit65:
@zhuyaliang 
Can you you confirm that meson.build files are missing?

zhuyaliang (approved):
No files were lost

raveit65:
@zhuyaliang 
But there is no meson.build file in /src/ui/ AND in src/meson.build we only list. Are you sure that a installation with meson with/from a tarball created with distcheck is working?

Issue #510, opened 2024-02-01
Engrampa not opening / listing 7z password compressed files when the option of password protect file names was chosen for create the file


acastan:
#### Expected behaviour

In a 7z password compressed files when the option of password protect file names was chosen for create the file, it is expected that when you open the file, Engrampa ask for the password, and if the password is fine the listing the content (files) inside the compressed .7z files

#### Actual behaviour

In Debian Testing, with Engrampa 1.26.1-3, since moved from p7zip to 7zip, nothing is happening. Engrampa don't ask for the password and don't show the content.

#### Steps to reproduce the behaviour

(1) In Debian Testing, with Engrampa choose some documents or a folder to compress. Choose extension .7z

(2) Choose password protection and choose a password

(3) and -IMPORTANT- choose the option of password protect also the filenames too 

Then try to open the file again with Engrampa and it didn't show any content and don't ask for password. But I still can work with the file from command line with command 7zip

#### MATE general version

Mate Desktop in Debian Testing is 1.26.2-1 , today 1 february 2024

#### Package version

Engrampa in Debian Testing is 1.26.1-3 , today 1 february 2024

#### Linux Distribution

Debian Testing 

#### Link to bugreport of your Distribution (requirement)

Excuse my poor English. I think I detect this issue happening when in Debian Testing moved from p7zip to 7zip weeks ago, and the package Engrampa was 1.26.1-2.

   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1061510

Then somebody report this bug that was solved and theoretically fixed in Engrampa 1.26.1-3 , but this patch don't fix the issue I am describing, so I tried to report the bug, but Debian Bug Reporting System was a little bit tricky to me, and finally I wrote to the Engrama Debian package maintainers that I found in:

   https://tracker.debian.org/pkg/engrampa

They are very nice and soon answer me that they don't have a fix for this and I should report the problem to the Engrampa Maintainers/Upstream Developers here.

zhuyaliang:
What is the version of ```7zip```

acastan:
The 7z comand version is 23.01 and the 7zip package version is 23.01+dfsg-8

acastan:
I tested now Engrampa 1.26.2 and the issue is still happening as with Engrampa 1.26.1

zhuyaliang:
I just finished my vacation, and I will solve this problem soon

acastan:
[...]

PR #513, opened 2024-02-26
fr-process.c: Setting error status

zhuyaliang:
Fix https://github.com/mate-desktop/engrampa/issues/510


zhuyaliang:
@cwendling Do you agree to merge? The issue has been fixed

cwendling (approved):
I am not sure of all the possible consequences this change has but:

* it seems to work fine (couldn't see any issue with it with limited testing)
* it looks conceptually sound (sets the field in all cases rather than a very limited subset, and other cases seem to be used)

So LGTM.

zhuyaliang:
Thank you for reviewing

Issue #515, opened 2024-03-15
Drag'n'drop blocks GUI


dirkf:
#### Expected behaviour

Following steps below, GUI should not be blocked, but just show a "not allowed" cursor while the drag-drop target is invalid.

Better, it should also be possible to drag-drop any selection from Engrampa to any target that accepts drag-drop from a file manager.   
 
#### Actual behaviour
Following the steps below, GUI shows a "drag copy" cursor when the invalid d'n'd target is hovered and a dialog appears `You don't have the right permissions to extract archives in the folder ""` but the normal cursor is never reactivated, despite the cursor responding to mouse movements, and the GUI remains blocked to mouse and keyboard actions. Releasing the dragging mouse button, mouse clicks, ESC, etc, have no effect.

Switching to a separate console session (c-a-F6), `pkill engrampa`, and switching back (c-a-F7) unblocks the GUI. Also, the dialog can be closed with <Alt+F4>, but that has no other effect. It's possible to switch to an previously opened Xfce Terminal running _bash_ with <Alt+Tab>: in the Terminal <Space> and <Enter>, and the plain arrow keys, are blocked but <Tab> and <Ctrl+j> can be used to invoke the `pkill` command.

Probably any entry in the PCManFM Places list behaves similarly. The Places list in Thunar doesn't. Also probably, the underlying issue could be a defect in PCManFM, but it shouldn't be possible for that to block the GUI. Eg, drag-dropping an item from Thunar to the PCManFM Places list doesn't have the same result (but/and that can be a valid action).

#### Steps to reproduce the behaviour
Installed:
MX Linux 23.2, kernel 6.1.0-18-686-pae, xfce4 4.18, xfwm4 4.18.0-1, Engrampa 1.26.0-1+deb12u2 (hurrah, no CSD!), pcmanfm 1.3.2.1

1. Plug in FAT32 USB stick.
2. Open PCManFM and select some folder on the local disk containing an archive file
3. Show the Places sidebar in pcmanfm: the USB device should be shown.
4. Open the archive in Engrampa
5. Select a file in the archive from the file list pane.
6. Drag the file towards the pcmanfm window: a file details strip becomes attached to the "drag" cursor. 
7. Move the drag cursor over the USB filesystem in the Places list
8. The "drag" cursor (for me it was a black one-fingered hand, but would depend on theming) instantly gets a "drag copy" + and a dialog (as above) opens.
[...]

PR #517, opened 2024-04-19
At the end of drag and drop, grab and release seat

zhuyaliang:
Fix https://github.com/mate-desktop/engrampa/issues/515

### DIFF
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -4161,6 +4161,25 @@ file_list_drag_begin (GtkWidget          *widget,
 }
 
 static void
+grab_widget (GtkWidget *widget)
+{
+
+    GdkWindow *window;
+    GdkDisplay *display;
+    GdkSeat *seat;
+
+    g_return_if_fail (widget != NULL);
+    window = gtk_widget_get_window (widget);
+    display = gdk_window_get_display (window);
+
+    seat = gdk_display_get_default_seat (display);
+    gdk_seat_grab (seat, window,
+                   GDK_SEAT_CAPABILITY_ALL, TRUE,
+                   NULL, NULL, NULL, NULL);
+    gdk_seat_ungrab (seat);
+}
+
+static void
 file_list_drag_end (GtkWidget      *widget,
 		    GdkDragContext *context,
 		    gpointer        data)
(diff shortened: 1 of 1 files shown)


zhuyaliang:
@cwendling Do you accept this repair method

zhuyaliang:
@cwendling Looking forward to your review

src/fr-window.c
```diff
@@ -4160,6 +4160,25 @@ file_list_drag_begin (GtkWidget          *widget,
...
+    GdkDisplay *display;
+    GdkSeat *seat;
+
+    g_return_if_fail (widget != NULL);
+    window = gtk_widget_get_window (widget);
+    display = gdk_window_get_display (window);
+
+    seat = gdk_display_get_default_seat (display);
+    gdk_seat_grab (seat, window,
+                   GDK_SEAT_CAPABILITY_ALL, TRUE,
```
cwendling, line 4177: That's probably sufficient
```suggestion
                   GDK_SEAT_CAPABILITY_ALL_POINTING, TRUE,
```

cwendling:
@zhuyaliang I don't like the "fix", but at least it seems to work.

The root issue seems to be inside *src/eggtreemultidnd.c*'s `egg_tree_multi_drag_motion_event()`, or possibly in how pcmanfm handles drag events.
IIUC it comes from pcmanfm requesting the drag data during drag motion rather than drag drop, but I'm not sure of the implications.
FWIW, File-Roller suffers from the same issue.

zhuyaliang:
@cwendling I suspect that the problem lies within ```pcmanfm```, and I will try to solve it again. If anyone needs this ```PR``` during this period, I will open it again
Thank you for your code review and feedback. Due to vacation, I was unable to reply to the message in a timely manner

zhuyaliang:
@cwendling I tried to fix the problem in [pcmanfm,(https://github.com/lxde/libfm/pull/103) but I'm not sure if this fix is correct，Please help review the code

zhuyaliang:
There are other issues with ```libfm```, which may not be the most perfect fix, but I don't want to invest too much time in ```libfm```. My focus is on ```mate desktop```

PR #519, opened 2024-04-26
Fix volume split button in "Compress" window

Mejioy:


lukefromdc:
This is actually two unrelated commits, both are bugfixes. I already know the first part works, will have to test the second

lukefromdc:
Both of these work fine, I was just able to confirm https://github.com/mate-desktop/engrampa/issues/522 and confirm that this fixes it. While I have an issue here with Engrampa not being able to see an encrypted file list, that is a separate issue. Engrampa can create archives with an encrypted file list, but at least on my Debian system with my build options it cannot openthem but file-roller can.

Merging this now as we need these two bugs fixed, hoping for a followup PR to deal with the encrypted file list case assuming that can be duplicated.

PR #529, opened 2025-03-11
ci: Add a GitHub Actions build

cwendling:


cwendling:
Based off/inspired by https://github.com/mate-desktop/mate-panel/pull/1415
This seems to work fairly well and is pretty fast.  Review welcome :)

lukefromdc:
I don't know much about how this works but I don't see any problems with doing this in general

PR #530, opened 2025-03-17
rar: Fix asking password for encrypted archives with RAR7

### LINKED ISSUE #526: Engrampa --extract-here not prompting for password
It no longer gives me a chance to enter a password. A gui box pops up informing me that it could not find the volume [passworded rar].r00 and that I used the incorrect password. (I never entered a password, it didn't give me the chance)

I'm running Mint Mate 22.1 Xia. It was working in Mint Mate 20.3 Una.

Is this a bug with Engrampa  or Mint 22.1?
The passworded files extract fine using unrar in a terminal but I prefer to use "extract here" with a right-click.

Image

cwendling:
RAR 7.0 outputs a different error when password is missing:

>
> Incorrect password for foo.rar
> Incorrect password for foo.rar
> Cannot find volume foo.r00

Whereas RAR 6.21 doesn't output the last 2 lines.

The error about volumes confuses the parser thinking there's a volume issue, which hides the incorrect password one.

Note that RAR 7.0 outputs this *most of the time*, but once in a while it'll also skip the last 2 lines.  Whether it is a race condition or not, it at least happens often enough with RAR 7.0 to consider this its normal behavior and handle it.

Fixes #526.

### DIFF
--- a/src/fr-command-rar.c
+++ b/src/fr-command-rar.c
@@ -697,7 +697,12 @@ fr_command_rar_handle_error (FrCommand   *comm,
 			volume_filename = g_path_get_basename (line + strlen ("Cannot find volume "));
 			error->gerror = g_error_new (FR_ERROR, error->status, _("Could not find the volume: %s"), volume_filename);
 			g_free (volume_filename);
-			break;
+
+			/* RAR7 complains about missing volume for incorrect passwords
+			 * as well, so don't make this a definite error in case we also
+			 * had an incorrect password one earlier */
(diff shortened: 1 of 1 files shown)


cwendling:
@chuggerguy could you test this?

Ideally we'd need a test with… something, probably multi-volume archives? to verify the modified handling doesn't cause issues there.  But I guess encrypted archives are more common anyway for this to be fairly low risk.

lukefromdc:
I only have unrar-free 1:0.3.1-1 installed, do not have RAR installed so cannot test this for anything more than
a successful build

PR #534, opened 2025-06-19
lha: Fix crash, parse lha list output correctly

fragglet:
The output from the `lha` list archive command is eclectic and inconsistent, and engrampa was actually crashing when opening certain archives (eg. lha_os2_208/h3_subdir.lzh in the Lhasa test suite). But this also fixes the list parsing more generally. With this change, engrampa now appears to successfully open all of the .lzh files in the Lhasa test suite correctly without crashing.

The first field in lha's list output either contains Unix permissions or an OS name in [brackets]. There was already hard-coded support for [MS-DOS], [generic], [unknown] and [Amiga], but other OS names were not handled properly. This is now fixed, with an approach that is mindful of the fact that the OS name can contain spaces.

Empty value (whitespace) in the UID/GID column is also handled correctly.  This was partially working because of the aforementioned special-casing, but now is fixed more generally.

Note this fix has already been merged to [file-roller](https://gitlab.gnome.org/GNOME/file-roller/-/merge_requests/138) and also submitted to lxqt-archiver (lxqt/lxqt-archiver#442).

### DIFF
--- a/src/fr-command-lha.c
+++ b/src/fr-command-lha.c
@@ -149,7 +149,7 @@ split_line_lha (char *line)
 	scan = eat_spaces (scan);
 	for (; i < N_FIELDS; i++) {
 		field_end = strchr (scan, ' ');
-		if (field_end == NULL) {
+		if (field_end == NULL || (i + 1) == N_FIELDS) {
 			field_end = scan + strlen(scan);
 		}
 


lukefromdc:
Note that I do not have test files for this, though looks from the CI that we don't have build issues

fragglet:
> Note that I do not have test files for this, though it looks from the CI that we don't have any build issues with this. I don't have any microsoft or apple systems to generate test files though

[Lhasa has a test suite](https://github.com/fragglet/lhasa/tree/master/test/archives) of archive files from different systems, feel free to try any of them out. I actually opened every .lzh file from the test suite to confirm it wasn't crashing any more (with a bash `for` loop to run `engrampa` on each file in turn).

lukefromdc:
Most of these I don't have a clue as to how to identify, h0_lh7.lzh opened fine.
To test this I need a known crasher

fragglet:
[Here's a screen recording showing the crash on my end](https://www.youtube.com/watch?v=9gWMRKboj6c). I haven't tested with Engrampa 1.28.2 but the crash happens with the immediate parent of my commit (commit 51a2a84bcd7bd10ff53cf87b58e4b3ec89011df1). This is on Debian testing (trixie), using Lhasa 0.4.0 from the same. Which lha are you using - Unix lha, Lhasa or jlha-utlis?

file-roller had the same crash, as did lxqt-archiver; as mentioned, I've submitted the same fix to those projects too.

fragglet:
Note: [the Travis CI build failed](https://app.travis-ci.com/github/mate-desktop/engrampa/builds/275532493) but I'm not sure why. Is the Travis build broken? It's been three months since the last commit was pushed to this repo.

lukefromdc:
Changes in Travis broke it here, and nobody currently at MATE knows how to fix it so we are replacing Travis with
github-actions for CI.

lukefromdc:
I have lhasa installed, but know very little of this, there are only a few archive types I normally use or even see 
outside of testing something. Maybe someone else here can weigh in

cwendling:
Looks pretty good to me, but I suggest 2 changes (see 2 top commits of https://github.com/mate-desktop/engrampa/commits/lha-fixes/):
* edb212b233f5f557006319cba85951d36da6d653: Use named field indices to make things a bit clearer
* 3431191e8b0f1304d8a39e40f42d1ac6357bb6f7: Fix support for filenames with spaces (basically, allow the 8th field to contain spaces). Fixes content of https://github.com/fragglet/lhasa/blob/master/test/archives/lha_os2_208/lfn.lzh for example.

Anyways, kudos for your quest on fixing this everywhere! And this does make the code a lot more robust, for now I don't see an obvious reason why it would crash even on random input (and I did try with throwing fairly random stuff at it, see #535, and no crash or particularly surprising behavior).

One detail is that the parser won't support a date-less entry, whereas apparently `lha` might output one (e.g. for the currently unsupported *pmarc124/pm0.pma*).  It's not a huge problem for now as it's not supported and it won't crash, but that could be improved.  However, as IIUC you're the Lhasa maintainer (?) maybe you could consider adding a machine-parsable output mode as an option? Verbose output seems a bit easier, but still not 100% consistent, and not the easiest to parse.  Anyway, that's potentially for the future, not for this PR :slightly_smiling_face: 

@fragglet if you feel my suggestions are worth it (IMO at least the filename with space support is important), feel free to add the commits to this PR, or tell me and I'll make a separate one after merging this.

@lukefromdc which distro are you using?  I reproduce the OP's crashes and related fixes on Debian Bookworm with Lhasa 0.3.1-4+b1 and 0.4.0 (from testing).

vkareh (approved):

PR #541, opened 2026-01-28
ci: restructure github actions to use the same structure for all repos

mbkma:


vkareh (commented):
This workflow is missing the `release.yml` file. Is that by design or oversight?

mbkma:
oversight, thanks :)

PR #543, opened 2026-02-26
Wayland DnD: offer text/uri-list with GVFS archive:// URIs

### LINKED ISSUE #423: Drag and Drop to extract does not work under wayland
Drag and Drop to extract does not work under wayland, same as
https://gitlab.gnome.org/GNOME/file-roller/-/issues/4

nyuutsu:
## Wayland DnD: offer text/uri-list with GVFS archive:// URIs

Fixes #423

Engrampa's drag-and-drop uses the XDS protocol (`XdndDirectSave0`), which
negotiates the drop destination via X11 window properties. These are no-ops
on Wayland, so dragging files out of engrampa into caja (or anything else)
just silently does nothing. The drop cursor shows up but no files come out.

This patch adds a `text/uri-list` target alongside the existing XDS one.
On Wayland, when the destination requests data after a drop, we hand it
GVFS `archive://` URIs instead of going through XDS. The GVFS archive
backend (`gvfsd-archive`) already knows how to expose archive contents
as a virtual filesystem, so the file manager can just copy files out via
normal GIO operations. Engrampa doesn't need to extract anything itself.

The archive gets mounted synchronously (nested main loop) at
`drag-data-get` time. There's a fast path via
`g_file_find_enclosing_mount` that skips the mount if it's already up
from a previous drag. If the mount fails (no GVFS, missing archive
backend, whatever), the handler returns FALSE and falls through to the
existing XDS path.

The `archive://` scheme embeds a URI in its host component, so the
archive path needs to be double-encoded (percent signs from the first
pass get encoded again). `build_archive_mount_uri()` handles this.

The X11 path is not touched. All the Wayland stuff is behind
`is_running_on_wayland()` checks, and the `gdk_property_change`/
`gdk_property_delete` calls in `file_list_drag_begin`/`drag_end` are
skipped on Wayland (where they were already no-ops anyway).

Tested on Gentoo with sway (wlroots), GTK 3.24, GVFS 1.56, engrampa
1.28.2, caja 1.28. Works for both the file list and folder tree views.

Caveats: this needs GVFS with the archive backend. That's a standard
MATE dependency (caja itself requires GVFS), but file managers that
don't use GIO won't know what to do with `archive://` URIs. The
synchronous mount blocks briefly (~50-200ms for local archives), which
is in the same ballpark as `gtk_dialog_run()` and well under typical
compositor DnD timeouts.

### DIFF
--- a/src/eggtreemultidnd.c
+++ b/src/eggtreemultidnd.c
@@ -32,7 +32,8 @@
 
 static GtkTargetEntry target_table[] = {
         { "XdndDirectSave0", 0, 0 },
-        { "XdndEngrampa0", 0, 1 }
+        { "XdndEngrampa0", 0, 1 },
+        { "text/uri-list", 0, 2 }
 };
 
 typedef struct
--- a/src/fr-window.c
+++ b/src/fr-window.c
@@ -28,6 +28,9 @@
 #include <gio/gio.h>
 #include <gdk/gdk.h>
 #include <gdk/gdkkeysyms.h>
+#ifdef GDK_WINDOWING_WAYLAND
+#include <gdk/gdkwayland.h>
+#endif
 #include <gdk-pixbuf/gdk-pixbuf.h>
 
 #include "actions.h"
@@ -85,6 +88,7 @@ static int             file_list_icon_size = 0;
 #define XDS_ATOM   gdk_atom_intern  ("XdndDirectSave0", FALSE)
 #define TEXT_ATOM  gdk_atom_intern  ("text/plain", FALSE)
 #define XFR_ATOM   gdk_atom_intern  ("XdndEngrampa0", FALSE)
+#define URI_LIST_ATOM gdk_atom_intern ("text/uri-list", FALSE)
 
 #define FR_CLIPBOARD (gdk_atom_intern_static_string ("_RNGRAMPA_SPECIAL_CLIPBOARD"))
 #define FR_SPECIAL_URI_LIST (gdk_atom_intern_static_string ("application/engrampa-uri-list"))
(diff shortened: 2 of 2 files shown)


vkareh (commented):
I wonder if we should make `text/uri-list` with `archive://` the default mechanism - not just on Wayland, but on X11 as well.

Basically amounts to removing the `is_running_on_wayland` checks (especially since a lot of X11 calls are no-ops on Wayland), then using `ensure_gvfs_archive_mounted` to determine whether you need to use `XdndDirectSave0` as a fallback.

Some edge cases here might be non-GIO file managers (e.g. Dolphin, PCManFM, etc.), but I don't even know if engrampa DnD works on those today anyway...

vkareh (commented):
Nice! This works really well on my side. It also fixes drag-and-drop from Engrampa to Nautilus, which otherwise doesn't work with XDS.

@lukefromdc can you test this on Wayland? Here's how to test this thing:
1. take any tar.gz file and open it in engrampa
2. drag and drop a file from inside engrampa into a file manager
3. that's it

I tested drag-and-drop into the following file managers using X11:
- caja (mate): works on `master` and on this PR
- thunar (xfce): works on `master` and this PR
- pcmanfm (lxqt): works on `master` and this PR
- nautilus (gnome): **does not** work on `master`, **works** on this PR! :+1:
- dolphin (kde): does not work on either `master` nor this PR :-1:

lukefromdc (commented):
Mostly works on wayland but I ran into one problem: dragging and dropping a FILE works fine, but on dragging and dropping a DIRECTORY, the contents of the directory get extracted and copied directly to the destination directory without being in the actual folder you tried to drag and drop into that destination directory. 

This could easily clog up someone's desktop with hundreds or even thousands of files that were supposed to be in their own subdirectory,

nyuutsu:
The good news is the don't-flatten-folders fix is easy. The bad news is a passworded archive I tested on misbehaved so now I need to go check a bunch of stuff. Reporting back soonish

svenstaro:
@nyuutsu not to put some undue stress on you but did you have some time to look at this?

nyuutsu:
Hello! No problem at all. Slipped my mind entirely. This week-ish I will attempt to fix my stuff

